PR #27, opened 2014-04-23
Trying to bind on a used port in Nanomsg will abort() the whole process.

tcr:
This is a misfeature in nanomsg. The Node module and should provide the option to mitigate this issue:

https://github.com/nanomsg/nanomsg/issues/235

Since it requires changes to the nanomsg FSM I'll propose a workaround in Node for the short-term. An async function which checks for a port's availability, cleans up, and throws an `error` event or otherwise succeeds.

Obviously there is time enough for a race condition but this might mitigate the majority of common use.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,6 +1,7 @@
 var nn = require('bindings')('node_nanomsg.node');
 
 var util = require('util');
+var url = require('url');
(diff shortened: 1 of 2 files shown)


nickdesaulniers:
looks like this has been fixed upstream.

PR #42, opened 2015-02-16
free nn_recv() buf

reqshark:
don't memory leak incoming messages onto the `HEAP`

• use the return value, `len` from `nn_recv()` to set the length of the `memcpy` to a node buffer that is later garbage collected by V8

### DIFF
--- a/src/node_nanomsg.cc
+++ b/src/node_nanomsg.cc
@@ -193,7 +193,7 @@ NAN_METHOD(Recv) {
 
     // Invoke nanomsg function.
     char *buf = NULL;
-    int len = nn_recv(s, &buf, NN_MSG, 0);
+    int len = nn_recv(s, &buf, NN_MSG, flags);
 
     v8::Local<v8::Value> h = NanNewBufferHandle(len);
     memcpy(node::Buffer::Data(h), buf, len);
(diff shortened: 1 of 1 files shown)


nickdesaulniers:
nice find!  deployed in v0.3.3!  Bent, did you use specific tooling to find this leak, or just from reading the code?

reqshark:
thanks, glad to contribute! i had done some local machine profiling with xcode, but I started wondering about it because I would leave the process on overnight remotely over a server, like you could do that with a couple pub/subs. i probably should have used dtrace but i dont have that on my server...

anyway i was sending small message sizes around 1K or less about every `25ms`. after like a day or two, leaving it on for at least 24 hours, the node process would begin to slow down, skipping messages more frequently and behaving sporadically just before ultimately crashing. Immediately upon restart it would jump right back to the original awesome perf you get out of it.. so i let it run a while doing that a few times just to confirm this wasn't a fluke. takes a few days a figure that out, and if you use daemonize or do auto-revive after exit then this behavior might not have been noticeable. 

although it seemed like the heap size was growing, sometimes memory allocations and garbage collection in V8 acts biological... moreover, i'm suspicious of what node or libuv thinks it's adding to the heap because memory leaks are sometimes not so traceable, even when you're following inodes around... well whether or not that was a memory leak, and if this update has anything to do with that, I think it'll be good to continue monitoring it.

nickdesaulniers:
neat to see people really stress testing this lib; long running processes!  I should spend some time with scan-build on this lib too.  Let me know if we still appear to be leaking memory, and thanks for the PR!

PR #43, opened 2015-02-18
Retrieve symbols on module load with nn_symbol

comick:
See http://nanomsg.org/development-bindings.html, this is the reccomanded way to retrieve symbols.

Not sure about:
EXPORT_CONSTANT(exports, EISCONN); // not in symbol.c, but is in nn.h
// Polling - these aren't in symbol.c but they are in nn.h
EXPORT_CONSTANT(exports, NN_POLLIN);
EXPORT_CONSTANT(exports, NN_POLLOUT);

why they were defined as symbol in node module?
They are not returned by nn_symbol and it seems they are not needed at all.
All tests pass.

### DIFF
--- a/src/node_nanomsg.cc
+++ b/src/node_nanomsg.cc
@@ -435,15 +435,13 @@ void InitAll(Handle<Object> exports) {
     EXPORT_METHOD(exports, Term);
 
     // Export symbols.
-    int index = 0;
-    int value;
-    while (true) {
-        const char *symbol_name = nn_symbol(index, &value);
+    for (int i = 0; ; ++i) {
+        int value;
+        const char *symbol_name = nn_symbol(i, &value);
         if (symbol_name == NULL) {
             break;
         }
         exports->Set(NanNew(symbol_name), NanNew<Number>(value));
-        index += 1;
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/node_nanomsg.cc
```diff
@@ -435,130 +434,17 @@ void InitAll(Handle<Object> exports) {
...
+    int index = 0;
+    int value;
+    while (true) {
+        const char *symbol_name = nn_symbol(index, &value);
+        if (symbol_name == NULL) {
+            break;
+        }
+        exports->Set(NanNew(symbol_name), NanNew<Number>(value));
+        index += 1;
+    }
```
nickdesaulniers: Can we turn this into a for-loop, use pre-increment on the loop variable, and put `int value` in the for-loop's block?

nickdesaulniers:
:+1: :+1: :+1: patches that remove more code than they add!  Just one fixer-upper and this is good to merge.

nickdesaulniers:
Looks good, thanks Michele.  Make sure you comment in the issue when you fix up code review nits; otherwise the maintainer doesn't get an email that the fixes were made. (or the maintainers GH settings are messed up)

nickdesaulniers:
published in v0.3.5

reqshark:
wow cool :+1:

comick:
@nickdesaulniers :+1:

PR #46, opened 2015-02-20
sockopt API

reqshark:
• convenience methods for `get/setsockopt()`

an interface to nanomsg socket options by way of:

`socket.option(param)` or `socket.option()`

passing a param sets that, while no param gets that.

*a few caveats will be referenced in forthcoming docs

it’s the simple interface i wrote in nanomsg.iojs covering:

• `tcpnodelay`
• `linger`
• `sndbuf`
• `rcvbuf`
• `sndtimeo`
• `rcvtimeo`
• `reconn`
• `maxreconn`
• `sndprio`
• `rcvprio`

cheers! here are the docs for this change: https://github.com/reqshark/nanomsg.iojs#sockettcpnodelayboolean

should we add those docs directly into the readme or maybe create a docs directory? I think there's going to be more documentation to follow this...

please comment and criticize the coding style: in the C/C++ I left `Getsockopt()` and `Setsockopt()` unchanged because i thought we should consider the differences and maybe merge them into something singular somehow.. also i dont want to break any tests yet!


lib/index.js
```diff
@@ -319,3 +333,134 @@ exports.socket = createSocket;
...
+  sndbuf          : nn.NN_SNDBUF,
+  rcvbuf          : nn.NN_RCVBUF,
+  sndtimeo        : nn.NN_SNDTIMEO,
+  rcvtimeo        : nn.NN_RCVTIMEO,
+  reconn          : nn.NN_RECONNECT_IVL,
+  maxreconn       : nn.NN_RECONNECT_IVL_MAX,
+  sndprio         : nn.NN_SNDPRIO,
+  tcpnodelay      : nn.NN_TCP_NODELAY
+}
+
```
reqshark: as far as style goes, should the `var sol` declaration be hoisted to the top of the file? btw i called it sol since it was named after socket level: `NN_SOL_SOCKET`

reqshark:
with these functions in place on the `this` value of the socket class, (alternatively you could use `module.exports` and the prototype declaration, etc.), but bear with pls, why i'd like to attach them on`this` is so that we can pass socket options at initialization, like:

### nano.socket(type, [options,])

```js
var nano = require('nanomsg')

var sub = nano.socket('sub', {
  tcpnodelay: true,
  rcvbuf: 1024
})

```

to get there from here, now we could just one-liner the [opts check inside our javascript socket class](https://github.com/nickdesaulniers/node-nanomsg/blob/master/lib/index.js#L291-L294) like:

```js
for(var sokopt in sol) if(opts.hasOwnProperty(sokopt)) this[sokopt](opts[sokopt])
```

`this[sokopt](opts[sokopt])` initializes any sockopt function with param passed by opts

if we do that then `var sol` object needs to be hoisted above [the Socket function declaration](https://github.com/nickdesaulniers/node-nanomsg/blob/master/lib/index.js#L10)

lib/index.js
```diff
@@ -85,6 +85,20 @@ function Socket (domain, type) {
...
+     this.linger     = linger
+     this.sndbuf     = sndbuf
+     this.rcvbuf     = rcvbuf
+     this.sndtimeo   = sndtimeo
+     this.rcvtimeo   = rcvtimeo
+     this.reconn     = reconn
+     this.maxreconn  = maxreconn
+     this.sndprio    = sndprio
+     this.rcvprio    = rcvprio
+     this.tcpnodelay = tcpnodelay
```
nickdesaulniers: I'll have to be more stringent about style throughout the code base, but please add semicolons.

lib/index.js
```diff
@@ -319,3 +333,134 @@ exports.socket = createSocket;
...
+var sol = {
+  linger          : nn.NN_LINGER,
+  sndbuf          : nn.NN_SNDBUF,
+  rcvbuf          : nn.NN_RCVBUF,
+  sndtimeo        : nn.NN_SNDTIMEO,
+  rcvtimeo        : nn.NN_RCVTIMEO,
+  reconn          : nn.NN_RECONNECT_IVL,
+  maxreconn       : nn.NN_RECONNECT_IVL_MAX,
+  sndprio         : nn.NN_SNDPRIO,
+  tcpnodelay      : nn.NN_TCP_NODELAY
```
nickdesaulniers: Let's add a trailing semi colon on the end of the last value; should there be more key/value pairs added later, no one has to touch your line.

lib/index.js
nickdesaulniers: I'm fine without explicit hoisting.  Maybe a comment pointing people to the man page to describe the socket options?

src/node_nanomsg.cc
```diff
@@ -23,6 +23,11 @@ using v8::Object;
 using v8::String;
 using v8::Value;
 
+#define ret NanReturnValue
+#define utf8 v8::String::Utf8Value
```
nickdesaulniers: ugh man, can we use c++11 safely yet in node add ons?
```c++
using utf8 = v8::String::Utf8Value;
```

src/node_nanomsg.cc
```diff
@@ -125,6 +130,41 @@ NAN_METHOD(Getsockopt) {
...
+
+NAN_METHOD(Getopt) {
+  NanScope();
+
+  int optval[64];
+  int option = args[2].integer;
+  size_t optsize = sizeof(optval);
+
+  //check if the function succeeds
+  if(nn_getsockopt(S, args[1].integer, option, optval, &optsize) == 0){
```
nickdesaulniers: Does this work correctly?  The third arg, `optval`, should be a `void*`.  Passing a `int [64]` is probably not what you meant to do?

lib/index.js
```diff
@@ -319,3 +333,134 @@ exports.socket = createSocket;
...
+}
+
+function linger(number){
+  if(number){
```
nickdesaulniers: Should we do any validation on the inputs?  What happens if you pass all of the functions strings?

nickdesaulniers:
While the syntax is certainly uglier for us, what do you think about making the methods getters/setters?
```javascript
for(var sokopt in sol) if(opts.hasOwnProperty(sokopt)) this[sokopt](opts[sokopt])
```
would become
```
for(var sokopt in sol) if(opts.hasOwnProperty(sokopt)) this[sokopt] = opts[sokopt];
```

nickdesaulniers:
> should we add those docs directly into the readme or maybe create a docs directory? I think there's going to be more documentation to follow this...

Let's put everything in the readme.  Want to add it to this PR, or make a followup issue?

src/node_nanomsg.cc
reqshark: haha yup good call, btw what do you think about that `#define S`, when socket is the first parameter of a libnanomsg call (which is almost every `nn_function()`).

personally, i was using `As<Number>()->IntegerValue()` which may do a 64 bit conversion on systems where int can be 64... i think it falls back to 32 otherwise, though i could be wrong

reqshark:
> Let's put everything in the readme. Want to add it to this PR, or make a followup issue?

i'll add that in a few minutes, with the rest of the changes from the comments

reqshark:
so can we do the opts initialization API? like:

### nano.socket(type, [options,])

where we pass opts that call respective `setsockopt()`methods?

nickdesaulniers:
> so can we do the opts initialization API? like:
It's a good idea.  Just do it :tm:

lib/index.js
reqshark: i guess... but isn't it kinda spicy without them?

src/node_nanomsg.cc
nickdesaulniers: > haha yup good call, btw what do you think about that #define S

I see that you mimic what's going on in the man page.  I don't know how I feel about 1 character macros, but it's fine.

> personally, i was using As<Number>()->IntegerValue() which may do a 64 bit conversion on systems where int can be 64... i think it falls back to 32 otherwise

I would think that `NanNew<Number>()` should handle the conversions correctly.

reqshark:
> While the syntax is certainly uglier for us, what do you think about making the methods getters/setters?
```js
this[sokopt] = opts[sokopt];
```

the point of that one-liner is just to call the respective sockopt setter.. firing a function call. i'm updating this and the comments, you'll see after I push the initializer stuff, I'll keep that in mind tho

lib/index.js
nickdesaulniers: maybe, I'm looking to use something like [semistandard](https://github.com/Flet/semistandard) though so can just write code however, and run the formatter during the tests.

lib/index.js
```diff
@@ -319,3 +333,134 @@ exports.socket = createSocket;
...
+    }
+  } else {
+    switch(nn.Getopt(this.binding, nn.NN_TCP, nn.NN_TCP_NODELAY)){
+      case 1: return 'tcp nodelay: on'
+      case 0: return 'tcp nodelay: off'
+      default:
+        throw new Error(nn.Err() + ': '+this.type+' nodelay@'+'getsockopt'+'\n')
+    }
+  }
+}
```
nickdesaulniers: Why does this function return strings of some custom format and not a bool?  Just make it explicit in the docs whether true/false means on/off.

src/node_nanomsg.cc
reqshark: ya but i'm talking about the unsigned int thing.

when declaring ints, i was using `As()->IntegerValue()` on `args[0]` as opposed to:

```c++
int s = args[0]->Uint32Value();
```

lib/index.js
nickdesaulniers: Oh, just saw all of these functions do.  Why not just return the value?  [It should be up to the consumer of the lib to log as they see fit.](http://www.linfo.org/rule_of_silence.html)

src/node_nanomsg.cc
reqshark: i wonder if that's why `rcvprio()` stopped working..

lib/index.js
nickdesaulniers: Also, should these be `= sol.linger; ...`?

reqshark: it's cool i'll add them, not using them makes me forget them when i switch to C/C++ sometimes anyway

lib/index.js
reqshark: should be fine, but i need to move the nn_socket() call out and up into https://github.com/nickdesaulniers/node-nanomsg/blob/master/lib/index.js#L291-L294 and then just give it to the binding as a parameter to 
```js
new Socket(domain, type);
```

src/node_nanomsg.cc
nickdesaulniers: `nn_setsockopt` and `nn_getsockopt` take a (signed) int as their first argument.  You shouldn't be using a precise type for int, instead defer to the host's definition of an int.

lib/index.js
reqshark: > Why does this function return a string of some custom format and not a bool? Just make it explicit in the docs whether true/false means on/off.

wow good call

nice article btw, here's a quote on topic from Ken Arnold:

> I think that the terseness of Unix programs is a central feature of the style. When your program's output becomes another's input, it should be easy to pick out the needed bits. And for people it is a human-factors necessity — important information should not be mixed in with verbosity about internal program behavior. If all displayed information is important, important information is easy to find.

src/node_nanomsg.cc
reqshark: nvm i guess C++11 extensions dont work yet in node

nickdesaulniers: It looks like you can pass cflags arguments to node-gyp, but note sure it's safe to bet on end users having c++11 support just yet.  I wonder what @trevnorris thinks?

lib/index.js
reqshark: trailing comma?

nickdesaulniers: yep, my bad

src/node_nanomsg.cc
trevnorris: It's a safe bet if compiling for io.js, since V8 requires C++11. But not safe in current node.js. the requirement for C++11 compatible compiler is why we didn't upgrade past 3.28.

reqshark:
just added the options initializer, so we can set any of those options as we create the socket. 

try it!

```js
var pub = nano.socket('pub')
var sub = nano.socket('sub', { tcpnodelay: true } )

console.log(pub.tcpnodelay()) //false
console.log(sub.tcpnodelay()) //true
```

also added some api docs to the readme.. just note that all the links to further down the document in the `nano.socket(type, [options,])` bullet points are linked based on the idea that the readme resides on master and not this branch

lib/index.js
```diff
@@ -290,7 +326,7 @@ util.inherits(Device, EventEmitter);
  */
 function createSocket (type, opts) {
     var domain = (opts || {}).raw ? nn.AF_SP_RAW : nn.AF_SP;
-    return new Socket(domain, type);
+    return new Socket(domain, type, opts);
```
nickdesaulniers: delete `var domain ...` and the first arg.  No need to pull it out of `opts` here, since we pass on `opts`.  Fix up the `Socket` constructor too to pull the domain out of `opts`.

lib/index.js
```diff
@@ -319,3 +355,114 @@ exports.socket = createSocket;
...
+}
+
+function maxreconn(number){
+  if(number){
+    if(setopt(this.binding, 'maxreconn', number) > -1) return true;
+    throw new Error(nn.Err() + ': ' + this.type + ' maxreconn@' + number +'\n');
+  } else {
+    return getopt(this.binding, 'maxreconn');
+  }
+}
```
nickdesaulniers: If someone wants to set maxreconn to `0`, this is going to act like a getter, not a setter.  Same for the others that expect a number.

test/sockoptapi.js
```diff
@@ -0,0 +1,49 @@
...
+  t.equal( sock.maxreconn(100000), true, 'sock.maxreconn(100000) sets: 100000ms')
+  t.equal( sock.maxreconn(), 100000, 'sock.maxreconn() gets: 100000')
+
+  //sndprio
+  t.equal( sock.sndprio(3), true, 'sock.sndprio(3) sets: 3 priority')
+  t.equal( sock.sndprio(), 3, 'sock.sndprio() gets: 3')
+
+  sock.close()
+  t.end()
+})
```
nickdesaulniers: semi colons

lib/index.js
reqshark: ya i was doing that but for some reason the `transform.js` test would time out.. tape would just stall forever wtf so i just left domain untouched. 

I'll restart my machine and pull down a fresh copy, not sure why domain opt would have anything to do with that test timing out..

nickdesaulniers: From simply removing a constructor argument?  You didn't rename the property did you?  see https://github.com/nickdesaulniers/node-nanomsg/blob/master/lib/index.js#L11

lib/index.js
```diff
@@ -319,3 +355,114 @@ exports.socket = createSocket;
...
+      if(nn.Setopt(this.binding, nn.NN_TCP, nn.NN_TCP_NODELAY, 0) > -1)
+        return false;
+      throw new Error(nn.Err() + ': '+this.type+' nodelay@'+'deactiving'+'\n');
+    }
+  } else {
+    switch(nn.Getopt(this.binding, nn.NN_TCP, nn.NN_TCP_NODELAY)){
+      case 1: return true;
+      case 0: return false;
+      default:
+        throw new Error(nn.Err() +': '+this.type+' nodelay@'+'getsockopt'+'\n');
```
nickdesaulniers: `nn.Err()` can replace two instances of `nn.Strerr(nn.Errno()))` in this file.
https://github.com/nickdesaulniers/node-nanomsg/blob/sockopts/lib/index.js#L134
https://github.com/nickdesaulniers/node-nanomsg/blob/sockopts/lib/index.js#L147
possibly:
https://github.com/nickdesaulniers/node-nanomsg/blob/sockopts/lib/index.js#L313

then we could remove them from the C++ code.

lib/index.js
reqshark: ya probably, sure. 

we definitely need to just pass opts, domain included, it's very ugly without doing that

lib/index.js
reqshark: yup better use `argments.length` instead of `number` since `0` is falsey

lib/index.js
reqshark: we dont have a domain param or var anymore so i got rid of the warning

lib/index.js
reqshark: the first two, definitely, but i had to think about the 3rd one: https://github.com/nickdesaulniers/node-nanomsg/blob/sockopts/lib/index.js#L313

since the callback here will fire on some libnanomsg error during an `nn_device()` operation, we can just use `nn.Err()` without needing to carry the param back up into JS

README.md
```diff
@@ -31,6 +31,187 @@ setTimeout(function () {
...
+#### `options`
+* `'raw'` *(Boolean, default: `false`)*: determines the domain of the socket. `AF_SP`, the default, creates a standard full-blown SP socket. `AF_SP_RAW` family sockets operate over internal network protocols and interfaces. Raw sockets omit the end-to-end functionality found in `AF_SP` sockets and thus can be used to implement intermediary devices in SP topologies, see [nanomsg docs](http://nanomsg.org/v0.5/nn_socket.3.html) or consult your man page entry `socket(2)` for more info.
+```js
+//ex. starting raw sockets
+nano.socket('bus', { raw: true } )
+```
+* `'tcpnodelay'` *(Boolean, default: `false`)*: see [`socket.tcpnodelay(boolean)`](https://github.com/nickdesaulniers/node-nanomsg#sockettcpnodelayboolean).
+* `'linger'` *(Number, default: `1000`)*: see [`socket.linger(duration)`](https://github.com/nickdesaulniers/node-nanomsg#socketlingerduration).
+* `'sndbuf'` *(Number, default: `128kB`)*: see [`socket.sndbuf(size)`](https://github.com/nickdesaulniers/node-nanomsg#socketsndbufsize).
+* `'rcvbuf'` *(Number, default: `128kB`)*: see [`socket.rcvbuf(size)`](https://github.com/nickdesaulniers/node-nanomsg#socketrcvbufsize).
```
nickdesaulniers: so if I said `1000`, would that be 1000B or 1000KB?

src/node_nanomsg.cc
```diff
@@ -125,6 +130,41 @@ NAN_METHOD(Getsockopt) {
...
+
+  int level = args[1].integer;
+  int option = args[2].integer;
+
+  if(option == NN_SOCKET_NAME){
+    utf8 str(args[3]);
+    ret(NanNew<Number>(nn_setsockopt(S, level, option, *str, str.length())));
+  } else {
+    int optval = args[3].integer;
+    ret(NanNew<Number>(nn_setsockopt(S, level, option, &optval, sizeof(optval))));
```
nickdesaulniers: One issue I just realized with this approach is that we always treat optval as a signed integer, but in fact `NN_LINGER` [can be negative](http://nanomsg.org/v0.2/nn_setsockopt.3.html)!
```
Negative value means infinite linger.
```

nickdesaulniers: Oops nvm. deleting last remark

src/node_nanomsg.cc
nickdesaulniers: This should be fixed before merging.

reqshark:
That Number check in JS is extra work for the JIT. No question. It will add up if u have to run a check on all inbound recv msgs.

If we adopt streams next, the Number type will cause an exception in the Readable base class, if you're in the middle of a buffer stream, and probably even object mode too

README.md
reqshark: 1000B

src/node_nanomsg.cc
nickdesaulniers: This does work, it's just using `63 * sizeof(int)` bytes more of the stack than it needs to.  It works because in C arrays decay into pointers when passed to functions.  So you're allocating space for 64 ints on the stack, then passing the address of the first element, which gets filled.  Since it's an array of ints, only the first element is ever populated.

If you think about the C ABI, it's not possible to represent 64 ints in a single register on today's processors.  That's why passing an array decays into a pointer.  A kind of syntactic sugar.  `nn_getsockopt` takes a `void*`, which works since `optval` decays into an `int*`.

A few lines down, there is no need to access `optval[0]` since we don't need `optval` to be an array of ints, just an int.

also, does `ret` actually return control flow from the function?

reqshark: interesting, especially about `int *`, and indeed, `ret` returns control. 

I like using `ret` because it feels cleaner than writing `NanReturnValue`

reqshark: let's just use one int on the register.

```c
  int optval[1];
  int option = args[2].integer;
  size_t optsize = sizeof(optval);

  //check if the function succeeds
  if(nn_getsockopt(S, args[1].integer, option, optval, &optsize) == 0){

    if(option == NN_SOCKET_NAME) ret(NanNew<String>((char *)optval));

    ret(NanNew<Number>(optval[0]));

```

reqshark:
can I check in my `Makefile`, take it off `.gitignore` list? Is there a reason it's being `.gitignored`?

the one i use eases build/test across different versions of node and OSs

src/node_nanomsg.cc
reqshark: `rcvprio()` doesn't work because `NN_RCVPRIO` symbol is not exported

it's supposed to be `9` after `NN_SOL_SOCKET` starts at `0`

this is what's being exported:

```js
_bindings: 
   { Socket: [Function],
     Close: [Function],
     Setsockopt: [Function],
     Getsockopt: [Function],
     Bind: [Function],
     Connect: [Function],
     Shutdown: [Function],
     Send: [Function],
     Recv: [Function],
     Errno: [Function],
     PollSendSocket: [Function],
     PollReceiveSocket: [Function],
     PollStop: [Function],
     DeviceWorker: [Function],
     SymbolInfo: [Function],
     Symbol: [Function],
     Term: [Function],
     Getopt: [Function],
     Setopt: [Function],
     Err: [Function],
     NN_NS_NAMESPACE: 0,
     NN_NS_VERSION: 1,
     NN_NS_DOMAIN: 2,
     NN_NS_TRANSPORT: 3,
     NN_NS_PROTOCOL: 4,
     NN_NS_OPTION_LEVEL: 5,
     NN_NS_SOCKET_OPTION: 6,
     NN_NS_TRANSPORT_OPTION: 7,
     NN_NS_OPTION_TYPE: 8,
     NN_NS_OPTION_UNIT: 9,
     NN_NS_FLAG: 10,
     NN_NS_ERROR: 11,
     NN_NS_LIMIT: 12,
     NN_TYPE_NONE: 0,
     NN_TYPE_INT: 1,
     NN_TYPE_STR: 2,
     NN_UNIT_NONE: 0,
     NN_UNIT_BYTES: 1,
     NN_UNIT_MILLISECONDS: 2,
     NN_UNIT_PRIORITY: 3,
     NN_UNIT_BOOLEAN: 4,
     NN_VERSION_CURRENT: 2,
     NN_VERSION_REVISION: 1,
     NN_VERSION_AGE: 2,
     AF_SP: 1,
     AF_SP_RAW: 2,
     NN_INPROC: -1,
     NN_IPC: -2,
     NN_TCP: -3,
     NN_PAIR: 16,
     NN_PUB: 32,
     NN_SUB: 33,
     NN_REP: 49,
     NN_REQ: 48,
     NN_PUSH: 80,
     NN_PULL: 81,
     NN_SURVEYOR: 96,
     NN_RESPONDENT: 97,
     NN_BUS: 112,
     NN_SOCKADDR_MAX: 128,
     NN_SOL_SOCKET: 0,
     NN_LINGER: 1,
     NN_SNDBUF: 2,
     NN_RCVBUF: 3,
     NN_SNDTIMEO: 4,
     NN_RCVTIMEO: 5,
     NN_RECONNECT_IVL: 6,
     NN_RECONNECT_IVL_MAX: 7,
     NN_SNDPRIO: 8,
     NN_SNDFD: 10,
     NN_RCVFD: 11,
     NN_DOMAIN: 12,
     NN_PROTOCOL: 13,
     NN_IPV4ONLY: 14,
     NN_SOCKET_NAME: 15,
     NN_SUB_SUBSCRIBE: 1,
     NN_SUB_UNSUBSCRIBE: 2,
     NN_REQ_RESEND_IVL: 1,
     NN_SURVEYOR_DEADLINE: 1,
     NN_TCP_NODELAY: 1,
     NN_DONTWAIT: 1,
     EADDRINUSE: 48,
     EADDRNOTAVAIL: 49,
     EAFNOSUPPORT: 47,
     EAGAIN: 35,
     EBADF: 9,
     ECONNREFUSED: 61,
     EFAULT: 14,
     EFSM: 156384766,
     EINPROGRESS: 36,
     EINTR: 4,
     EINVAL: 22,
     EMFILE: 24,
     ENAMETOOLONG: 63,
     ENETDOWN: 50,
     ENOBUFS: 55,
     ENODEV: 19,
     ENOMEM: 12,
     ENOPROTOOPT: 42,
     ENOTSOCK: 38,
     ENOTSUP: 45,
     EPROTO: 100,
     EPROTONOSUPPORT: 43,
     ETERM: 156384765,
     ETIMEDOUT: 60,
     EACCES: 13,
     ECONNABORTED: 53,
     ECONNRESET: 54,
     EHOSTUNREACH: 65,
     EMSGSIZE: 40,
     ENETRESET: 52,
     ENETUNREACH: 51,
     ENOTCONN: 57,
     path: '/Users/bent/routekickr/node-nanomsg/build/Release/node_nanomsg.node' },
  Socket: 

```

nickdesaulniers: [We were recently mucking with the exported symbols](https://github.com/nickdesaulniers/node-nanomsg/pull/43), so it's possible we missed one.

This function has bigger issues, do to the fact that set/getsocketopts can work on a single int, or a `char[]` when the level is `NN_SUB` and the option is `NN_SUB_SUBSCRIBE`.  Please see the examples sections from [set](http://nanomsg.org/v0.2/nn_setsockopt.3.html) and [get](http://nanomsg.org/v0.2/nn_getsockopt.3.html).  See also [usage](https://github.com/nanomsg/nanomsg/search?utf8=%E2%9C%93&q=nn_getsockopt) within the lib.

@sustrik, under what conditions does `nn_getsocketopts` expect to be passed a `char[]`?

README.md
nickdesaulniers: Let's make sure to note in the docs then that the units are bytes, `size in bytes`.

reqshark:
yo @nickdesaulniers can I check my Makefile into this repo and take `Makefile` off `.gitnore`?

reqshark:
updates to commit 9fa33004c274f95cb68c367302e3abd86ee52468:

> These should be added on the prototype, no need to have unique copies per Socket instance.

done.

> I wonder what it would look like without call

behold.

> you could probably switch the order of the arguments

done.

lib/index.js
```diff
@@ -3,16 +3,32 @@ var nn = require('bindings')('node_nanomsg.node');
...
+var sol = {
+  linger          : nn.NN_LINGER,
+  sndbuf          : nn.NN_SNDBUF,
+  rcvbuf          : nn.NN_RCVBUF,
+  sndtimeo        : nn.NN_SNDTIMEO,
+  rcvtimeo        : nn.NN_RCVTIMEO,
+  reconn          : nn.NN_RECONNECT_IVL,
+  maxreconn       : nn.NN_RECONNECT_IVL_MAX,
+  sndprio         : nn.NN_SNDPRIO,
+  rcvprio         : 9,
```
nickdesaulniers: I'm not sure how I feel about committing this in this PR. I think in a follow up, I'd like to have a test case that shows it's broken and a comment
```
// TODO: Issue #50
```

lib/index.js
```diff
@@ -85,6 +101,12 @@ function Socket (domain, type) {
     this.binding = nn.Socket(this.af_domain, this.protocol);
     this.queue = [];
 
+    opts = opts || {}
```
nickdesaulniers: If this was moved above line 28, then we wouldn't need `(opts || {})` on line 29.

lib/index.js
```diff
@@ -286,11 +308,58 @@ function Device (sock1,sock2) {
...
+ */
+Socket.prototype.linger     = function (o) { return opt('linger',    o ); }
+Socket.prototype.sndbuf     = function (o) { return opt('sndbuf',    o ); }
+Socket.prototype.rcvbuf     = function (o) { return opt('rcvbuf',    o ); }
+Socket.prototype.sndtimeo   = function (o) { return opt('sndtimeo',  o ); }
+Socket.prototype.rcvtimeo   = function (o) { return opt('rcvtimeo',  o ); }
+Socket.prototype.reconn     = function (o) { return opt('reconn',    o ); }
+Socket.prototype.maxreconn  = function (o) { return opt('maxreconn', o ); }
+Socket.prototype.sndprio    = function (o) { return opt('sndprio',   o ); }
+Socket.prototype.rcvprio    = function (o) { return opt('rcvprio',   o ); }
```
nickdesaulniers: Nice, now we can do:
```javascript
Socket.prototype.linger = opt.bind('linger');
...
```

partial function application baby!

nickdesaulniers:
This is turning out to be a great addition so far!  Thanks for all of your hard work, Bent!

reqshark:
haha thanks for that, my pleasure, Nick! Just saw your latest comments, i'll fix/update those this evening

reqshark:
@nickdesaulniers I think that should just about do it, anything else?

Issue #29, opened 2014-04-30
streams2 api?


jandre:
Hi,

Have you thought about a streams api for nanomsg?   

I'd like to be able to implement nanomsg with some sort of backpressure support if my downstream node stuff, for example, is too busy to actually process PULL messages coming up from my nanomsg pipeline.   The EventEmitter callback-style isn't *really* ideal since I assume it will just keep accepting messages as long as I'm still listening and has the potential to just queue up callbacks in memory, if, say, I'm inserting messages into the db on the backend and the db insert can't keep up.

I'm looking at implementing one myself if you haven't thought about it already, or if you have and have any pointers or thoughts, would love to hear them!

Best,
Jen

reqshark:
a wonderful idea and we ought to wholesale replace EventEmitter with the `Duplex` base class.

just get rid [the call to `emit('message')`](https://github.com/reqshark/node-nanomsg/blob/streams/lib/index.js#L149)

the API change would be like `.on('data', callback)` instead of `.on('message', callback)`

`data` is a few characters shorter than `message` so it seems easier. plus you get a node stream.

keep the call to send the same, no need for any `write()` method, yet all sender type sockets could be piped to from any source. by adding a `_write()` method on the prototype, the `Duplex` base class takes care of handing the writing for us, check it out:

https://github.com/reqshark/node-nanomsg/commit/9ea49b3864a5e0f877055bd7bbb83aeb0c00c23d

what do people think?

PR #53, opened 2015-02-25
nanomsg streams: subclass Socket stream from Duplex base class

reqshark:
PR builds directly on @tcr efforts, leveraging initial work and direction taken in https://github.com/nickdesaulniers/node-nanomsg/pull/30: 
* the Readable Stream's `_read(n){}` implementation is no different here.
* furthers #29, implementing a `Writeable`

#### other notable characteristics:
* revised program interface for all send and receive operations
* removes EventEmitter
* new API:
  - full blown Duplex interface: Writeable and Readable streams
  - from `.on('message', callback)` to `.on('data', callback)`
  - pipe all endpoints together from chainable  `pipe()` methods
  - socket’s `pipe()` method is essentially a flexible `zmq_proxy()` or `nn_device()`
  - io.js streams are domain, protocol, and transport agnostic
  - combine sockets in new ways and connect modules like [websocket stream](https://github.com/maxogden/websocket-stream)

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ pub.bind(addr);
 sub.connect(addr);
 
 sub.on('message', function (buf) {
-	console.log(buf.toString());
+	console.log(String(buf));
 	pub.close();
 	sub.close();
 });
@@ -94,6 +94,51 @@ socket.connect('tcp://127.0.0.1:5555');
 
 *(Function, param: Function)*: Closes the socket. Any buffered inbound messages that were not yet received by the application will be discarded. The nanomsg library will try to deliver any outstanding outbound messages for the time specified by `linger`.
 
+## sending and receiving: writable and readable
+
+### socket.send(msg)
+*(Function, param: String or Buffer)*: send a message.
+
+```js
+socket.send('hello from nanømsg!');
+```
+
+`send(msg)` is automatically invoked during `Writable` consumption of some other `Readable` stream. In that case a `Writable`'s `pipe()` method can be used to transmit across readable data sources. See [example for more detail](examples/writablepipe.js). The flow of data distributes to endpoint(s) determined by the particular socket type.
+
+```js
+var source = require('fs').createReadStream('filename.ext');
+
+source.pipe(socket); //sends each chunk as a msg to socket's particular endpoint
+```
+
+### socket.on('data', callback)
+*(Function, param order: String, Function)*: The `Readable` stream's `on()` function is an event listener that emits `'data'` events. To receive messages, pass the string `'data'` followed a callback containing a single data parameter.
+
+```js
+// the default inbound message is a node buffer
+// setEncoding sets the message type, use utf8 to receive strings instead.
+socket.setEncoding('utf8');
+
+socket.on('data', function (msg) {
+  console.log(msg); //'hello from nanømsg!'
+});
+```
+
+The readable stream's `data` event is automatically invoked when piped to a `Writable` or `Transform` consumer stream. See [example for more detail](examples/transforms.js). Here `msgprocessor` is a transform you could pipe to a writable or the next transform:
+
+```js
+var through = require('through');
+
+var msgprocessor = through(function(msg){
+  var str = msg; //'hello from nanømsg'
+  this.queue(str + ' and cheers!');
+});
+
+socket.pipe(msgprocessor); //msg transformed to: 'hello from nanømsg and cheers!'
+```
+
+## subscription api
+
 ### socket.chan(Array)
 
 *(Function, param: Array of Strings, default: `['']`)*: Allows for sub sockets
@@ -110,6 +155,8 @@ multiple strings and all will be unfiltered.
 
 If you unsubscribe from the default channel, `''`, without subscribing to any
 new channels, your sub socket will stop receiving messages.
+
+## sockopt api
 
 ### socket.tcpnodelay(boolean)
 
--- a/examples/writablepipe.js
+++ b/examples/writablepipe.js
@@ -1,0 +1,16 @@
+var nano = require('..')
+var pub   = nano.socket('pub'),       push  = nano.socket('push');
+var sub   = nano.socket('sub'),       pull  = nano.socket('pull');
+pub.bind('tcp://127.0.0.1:3333');     push.bind('tcp://127.0.0.1:4444');
+sub.connect('tcp://127.0.0.1:3333');  pull.connect('tcp://127.0.0.1:4444');
+
+// setEncoding formats inbound message type
+sub.setEncoding('utf8');
+
+sub.on('data', function (msg) {
+  console.log(msg); //hello from a push socket!
+})
+
+pull.pipe(pub); //pipe readable sockets to any writable socket or stream
+
+setInterval( function(){ push.send('hello from a push socket!') }, 100 );
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,7 +1,7 @@
-var nn = require('bindings')('node_nanomsg.node');
-
-var util = require('util');
-var EventEmitter = require('events').EventEmitter;
+var nn            = require('bindings')('node_nanomsg.node');
+var util          = require('util');
+var EventEmitter  = require('events').EventEmitter;
+var Duplex        = require('stream').Duplex;
 
 /**
  * generic socket-level NN_SOL_SOCKET options
@@ -121,12 +121,21 @@ function Socket (type, opts) {
   }
 
   /* start listening for inbound messages */
-  if(this.af_domain === nn.AF_SP) {
-    if (this.receiver) this._startPollReceive();
+  if (this.af_domain === nn.AF_SP && this.receiver) {
+    this._startPollReceive();
   }
+
+  if (!opts.hasOwnProperty('objectMode')) {
+    // reduce the highwatermark from 16K msgs to 16msgs
+    // would be more useful if we have a backpressure mechanism
+    opts.objectMode = true;
+  }
+
+  Duplex.call(this, opts);
+
 }
 
-util.inherits(Socket, EventEmitter);
+util.inherits(Socket, Duplex);
 
 Socket.prototype._protect = function (ret, unwind) {
     if(ret < 0) {
@@ -166,11 +175,11 @@ Socket.prototype._send = function (msg, flags) {
 };
 
 Socket.prototype._receive = function () {
-    if (this.closed) return;
+  if (this.closed) return null;
 
-    var msg = nn.Recv(this.binding, 0);
+  var msg = nn.Recv(this.binding, 0);
 
-    if(this.type == 'surveyor') {
+  if(this.type == 'surveyor') {
         if(msg < 0 && nn.Errno() == nn.EFSM) {
             this._stopPollSend();
             this._stopPollReceive();
@@ -178,10 +187,10 @@ Socket.prototype._receive = function () {
             return;
         }
     }
-    if (msg == -1) return;
+  if (msg == -1) return null;
 
-    if (this.restore && typeof this.restore === 'function') msg = this.restore(msg);
-    this.emit('message', msg);
+  if (this.restore && typeof this.restore === 'function') msg = this.restore(msg);
+  return msg;
 };
 
 Socket.prototype._startPollSend = function () {
@@ -193,11 +202,11 @@ Socket.prototype._startPollSend = function () {
 }
 
 Socket.prototype._startPollReceive = function () {
-    if (!this._pollReceive) {
-        this._pollReceive = nn.PollReceiveSocket(this.binding, function (events) {
-            if (events) this._receive();
-        }.bind(this));
-    }
+  if (!this._pollReceive) {
+    this._pollReceive = nn.PollReceiveSocket(this.binding, function (events) {
+      if (events) this.push(this._receive());
+    }.bind(this));
+  }
 }
 
 Socket.prototype._stopPollSend = function () {
@@ -220,6 +229,13 @@ Socket.prototype._register = function(chan){
     this.emit('error', new Error( nn.Err() + ' : ' + chan));
   }
 };
+
+Socket.prototype._write = function(buf, _, cb){
+  this.send(buf);
+  cb();
+}
+
+Socket.prototype._read = function (n) {}
 
 /**
  * Socket API
@@ -307,7 +323,7 @@ Socket.prototype.survey = function (buf, callback) {
         this.removeListener('message', listener);
         callback(responses);
     })
-    this.on('message', listener)
+    this.on('data', listener)
     this.send(buf);
 }
 
@@ -338,6 +354,8 @@ function Device (sock1,sock2) {
     } else {
         throw new Error('expected at least one Socket argument');
     }
+
+    EventEmitter.call(this);
 }
 
 util.inherits(Device, EventEmitter);
--- a/perf/local_lat.js
+++ b/perf/local_lat.js
@@ -19,7 +19,7 @@ rc = s.bind(bind_to);
 assert(rc >= 0);
 
 var i = 0;
-s.on('message', function (data) {
+s.on('data', function (data) {
     assert.equal(data.length, sz);
     var nbytes = s.send(data);
     assert.equal(nbytes, sz);
--- a/test/streams.js
+++ b/test/streams.js
@@ -1,0 +1,36 @@
+var nano = require('..');
+var test = require('tape');
+
+test('pipe a thousand msgs between incompatible socket types', function(t){
+
+  var sent = 0, recv = 0;
+
+  var pub = nano.socket('pub', { tcpnodelay: true });
+  var sub = nano.socket('sub', { tcpnodelay: true });
+
+  var push = nano.socket('push');
+  var pull = nano.socket('pull');
+  pull.setEncoding('utf8'); //should also be able to do in `socket(type, opts)`
+
+  pub.bind('tcp://127.0.0.1:64999');
+  sub.connect('tcp://127.0.0.1:64999');
+
+  pull.bind('tcp://127.0.0.1:65000');
+  push.connect('tcp://127.0.0.1:65000');
+
+  sub.pipe(push);
+  pull.on('data', pullsocket);
+
+  while(sent++ < 1001) pub.send('hello from nanomsg pub socket!');
+
+  function pullsocket(msg){
+    if(recv++ > 999){
+      t.equal( msg, 'hello from nanomsg pub socket!', 'piped a pub/pull combo');
+      pub.close();
+      push.close();
+      pull.close();
+      sub.close();
+      t.end();
+    }
+  }
+});
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ perf:
 	node perf/local_thr.js tcp://127.0.0.1:5556 1 100000& node perf/remote_thr.js tcp://127.0.0.1:5556 1 100000 && wait
 
 bench:
-	node perf/local_lat.js tcp://127.0.0.1:5555 10 1000& node perf/remote_lat.js tcp://127.0.0.1:5555 10 1000 && wait
+	node perf/local_lat.js tcp://127.0.0.1:5555 10 10000& node perf/remote_lat.js tcp://127.0.0.1:5555 10 10000 && wait
 	node perf/local_thr.js tcp://127.0.0.1:5556 10 100000& node perf/remote_thr.js tcp://127.0.0.1:5556 10 100000 && wait
 
 full: clean ALL test
--- a/examples/multisub.js
+++ b/examples/multisub.js
@@ -11,20 +11,20 @@ sub1.connect(addr);
 sub2.connect(addr);
 sub3.connect(addr);
 
-sub1.on('message', function (buf) {
-    console.log("sub1 got: %s", buf.toString());
+sub1.on('data', function (str) {
+  console.log('sub1 got: %s', str);
+  sub1.close();
 });
-
-sub2.on('message', function (buf) {
-    console.log("sub2 got: %s", buf.toString());
+sub2.on('data', function (str) {
+  console.log('sub2 got: %s', str);
+  sub2.close();
 });
-
-sub3.on('message', function (buf) {
-    console.log("sub3 got: %s", buf.toString());
+sub3.on('data', function (str) {
+  console.log('sub2 got: %s', str);
+  sub3.close();
 });
 
 setTimeout(function () {
-    console.log("PUBLISHING...");
-    pub.send("Hello from nanomsg!");
+  console.log("PUBLISHING...");
+  pub.send("Hello from nanomsg!");
 }, 100);
-
--- a/examples/pipeline.js
+++ b/examples/pipeline.js
@@ -1,18 +1,16 @@
-var nano = require('../');
+var nano = require('..');
 
-var pull = nano.socket('pull');
+var pull = nano.socket('pull', {encoding:'utf8'} );
 var push = nano.socket('push');
 
-var addr = 'tcp://127.0.0.1:7789'
-pull.bind(addr);
-push.connect(addr);
+var addr = 'tcp://127.0.0.1:7789';
+pull.connect(addr);
+push.bind(addr);
 
-pull.on('message', function (buf) {
-	console.log(buf.toString());
-	pull.close();
-	push.close();
-});
+pull.on('data', console.log);
 
-setTimeout(function () {
-	push.send("Hello from nanomsg!");
-}, 100);
+setInterval( send, 100 );
+
+function send(){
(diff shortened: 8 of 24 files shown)


reqshark:
#### PR status: WIP

nickdesaulniers:
It might be worthwhile to reach out to someone in node core to review that our stream API is implemented correctly (in regards to back pressure).

reqshark:
well if we're producing some readable consumable it needs to moderate `nn_recv()` i/o

reqshark:
this last commit is a good stumbling step closer to a `push/pull` socket stream with built in backpressure. The tests are failing but i still felt like i was at a commit point because all socket types are covered except `respondent`/`surveyor` 

### where we're at:

When the stream is ready to `_read()` the next chunk, we're calling `this._receive()`, the pull socket's `Readable()` data source.  

`_receive()` method is our wrapper around the binding's access to `nn_recv()` operation. 

`nn_recv()` does a blocking read on the system's socket which returns a blob of data back up to V8 and nodejs, so we can work with a message buffer/string. It should be able to do upwards of a few hundred thousand per second without much special hardware, so I've tried to make the check for `pull` first before entering these I/O routines:
```js
      if (type == 'pull'){
        this._startPullReceive();
        this._read = function (n) {
          return this.push(this._receive());
        }
      } else {
        this._startPollReceive();
        this._read = function (n) {}
      }
```
*disclaimer: next description is limited by my understanding and assumptions about the interworkings of nanomsg, so it is possible this is a misrepresentation:*

A streaming backpressure scenario is now possible because the next inbound msg *cannot* arrive at the the socket until `nn_recv()` completes and (usock aio summary) switches socket state from:
#### state 1: *socket is ready for I/O, awaiting `nn_recv()` ABI call*
#### state 2: *socket is ready for inbound data, open to receive (nn_recv not possible)*

there is more involved in the nanomsg internal socket state at any given time, but those are the two handles (events) by which libuv code in our binding hooks Node's runloop with async data callbacks in JS. @tcr wrote a lot of that amazing stuff. Anyway, [Refer to image below and the other diagrams](http://nanomsg.org/diagrams.html) if interested, but that's basically how i see it. please correct me if i am wrong.

bottom line is: `pipe()` pull socket streams, call `pause()`, `resume()` and since `pull` socket's nn_recv() is moderated by a Readable stream's `_read()` implementation, any other node stream techniques should be available to throttle inbound `pull` message reception at the socket. So hopefully that's doable w/out writing too much intermediary queue stuff or bloated broker type code.

nn_usock_handler

asafyish:
Hi,

Any progress on that ? We are transferring large messages, and causing a large memory spike, with streams with can just stream the processed message, faster and with less memory.

Hope it lands soon ! :)

reqshark:
hi @asafyish, thanks for the ping on this I appreciate it. I was planning to circle back here shortly. 

Some core API changes I'd to see landed first:

- [ ] `socket.shutdown(addr)`
- [ ] `updates to travis`, to easily test the next commits
- [ ] pubsub channel API

the only reason I'm waiting for these to land first is because there's a merge conflict right now with this branch and it would easier to deal with after some of the API in flux settles down. probably next week.

asafyish:
How does streaming is going to work ?
Plain old streams have no boundaries between messages, they are just a bytestream.
Nanomsg api is giving as a "message" abstraction.
How the two are combined ?

reqshark:
Hey Asaf, You can pipe a source stream like pub or push to a consumeable and it automatically takes care of socket.send(msg). Msg is the stream's data.  Play around with [nanomsg.iojs](https://github.com/reqshark/nanomsg.iojs) if u want to get a feel for it before it's ready here

reqshark:
> Especially, How is the new streaming API going to handle large messages ? (in the range of a few hundreds mb's).

taking care of breaking up large messages could be out of scope here. 

honestly i havent even thought of that.

the thing is you get better throughput on large messages, but if memory is an issue, then you might need something that opens a special channel or socket for handling that and that's where SNDBUF sockopt should also be used.

reqshark:
as far as docs go I was thinking:

## sending and receiving: writeable and readable

### socket.send(msg)
*(Function, param: String or Buffer)*: equivalent to the `socket.send()` in [node.zeromq](https://github.com/JustinTulloss/zeromq.node).

```js
socket.send('hello from nanømsg!')
```

`send(msg)` is automatically invoked during `Writeable` consumption of some other `Readable` stream. In that case a `pipe()` method can be used to transmit from a readable data source. The flow of data distributes to endpoint(s) determined by the particular socket type.

```js
var fs = require('fs')
var source = fs.createReadStream(__dirname + 'filename.ext')

source.pipe(socket) //sends each chunk as a msg to socket's particular endpoint
```

### socket.on(data, callback)
*(Function, param order: String, Function)*: The `Readable` stream's `on()` function is an event listener registered with the `nanomsg c lib` that emits `'data'` events. To receive messages, pass the string `'data'` followed a callback containing a single data parameter.

```js
socket.on('data', function (msg) {
  console.log(String(msg)) //'hello from nanømsg!'
})
```

the readable stream's `data` event is automatically invoked when piped to a `Writeable` or `Transform` consumer stream. Here `msgprocessor` is a transform you could pipe to a writeable or the next transform:

```js
var through = require('through')

var msgprocessor = through(function(msg){
  var str = String(msg); console.log(str) //'hello from nanømsg!'
  this.queue(str + ' and cheers from nanomsg.iojs!')
})

socket.pipe(msgprocessor)

**Comment by reqshark** [comment]
@nickdesaulniers streams PR is ready for review!

I've been trying to set up a pull socket backpressure mechanism that could limit inbound message reception via the Readable `_read(n)` implementation. Unfortunately it seems the `_read(n)` method I set on the Socket prototype will only work successfully for a `pull` socket and that same implementation doesn't fly with other receiver socket types. 

As soon as I try to overwrite this prototype method conditionally for different receiver socket types, the whole message transport seems to break down.

We can probably skirt around the problem by building a unique prototype for unique socket types. This way we could write stable, custom methods appropriately crafted for the quirks of different sockets, which I think was the direction taken for surveyor at one point. See https://github.com/nickdesaulniers/node-nanomsg/blob/f08bdd09aee5ebafe1afbc98f231c70866fdcc04/lib/index.js#L295-L312 also note the amount of conditional checks we're making for type-specific differences in the `receive()` method: https://github.com/nickdesaulniers/node-nanomsg/blob/f08bdd09aee5ebafe1afbc98f231c70866fdcc04/lib/index.js#L168-L185 

In the meantime, we can keep the single Socket class and still have a stable Streams API by using a more generic `_read(n)` implementation. However, that means we need to compromise on the `pull` socket's backpressure mechanism. For now at least, that might be the right thing to do since it gives us a consistent Streams interface across all socket types.

I've updated the readme and added some documentation for streams. 

And all tests are green! :rocket:

**Comment by reqshark** [comment]
looking at performance I'm thinking to enable streams as an option at socket initialization.

send hasn't changed, so Readable's `.on('data')` seems to have a noticeable impact on performance.

I can't imagine perf would improve by piping the `Readable` to a `Transform` or `Writable`

**Comment by nickdesaulniers** [comment]
woah awesome thank you!  I'll try to get this reviewed this week.  Neat stuff.

File: lib/index.js [hunk 1]
```diff
@@ -1,7 +1,7 @@
 var nn = require('bindings')('node_nanomsg.node');
 
 var util = require('util');
-var EventEmitter = require('events').EventEmitter;
```
nickdesaulniers: Let's keep this here, since we end up requiring it twice down below.  I _know_ requires are cached, but let's keep imports at the top of the file, and remove the two requires down below.

perf/local_lat.js
```diff
@@ -19,7 +19,7 @@ rc = s.bind(bind_to);
 assert(rc >= 0);
 
 var i = 0;
-s.on('message', function (data) {
+s.on('data', function (data) {
```
nickdesaulniers: It looks like we'll be introducing breaking changes with this then, since users of this lib now have to listen for on data rather than on message?

test/streams.js
```diff
@@ -0,0 +1,36 @@
...
+test('pipe a thousand msgs between incompatible socket types', function(t){
+
+  var sent = 0, recv = 0;
+
+  var pub = nano.socket('pub', { tcpnodelay: true });
+  var sub = nano.socket('sub', { tcpnodelay: true });
+
+  var push = nano.socket('push');
+  var pull = nano.socket('pull');
+  pull.setEncoding('utf8'); //should also be able to do in `socket(type, opts)`
```
nickdesaulniers: What happens if we don't set the encoding?  We get a buffer instead of a string?  Did we have to do that before this patch?

lib/index.js
```diff
@@ -121,12 +121,19 @@ function Socket (type, opts) {
...
   /* start listening for inbound messages */
-  if(this.af_domain === nn.AF_SP) {
-    if (this.receiver) this._startPollReceive();
+  if (this.af_domain === nn.AF_SP && this.receiver) {
+    this._startPollReceive();
   }
+
+  if (!opts.hasOwnProperty('objectMode')) {
+    opts.objectMode = true;
+  }
```
nickdesaulniers: What does this property do?

lib/index.js
```diff
@@ -121,12 +121,19 @@ function Socket (type, opts) {
...
+
+  Duplex.call(this, opts);
```
nickdesaulniers: Does the duplex constructor really need `opts`?

lib/index.js
reqshark: oh good call

lib/index.js
nickdesaulniers: Ah, is this for the `setEncoding` thing?

perf/local_lat.js
reqshark: well yes, but i was thinking about activating the new API with an initializing flag option given to the socket since there's a performance hit involved in switching to the streams API

lib/index.js
```diff
@@ -178,10 +185,10 @@ Socket.prototype._receive = function () {
             return;
         }
     }
-    if (msg == -1) return;
+  if (msg == -1) return '';
```
nickdesaulniers: Why do we change the function signature to return `null` vs `empty string`?  Why would have to differentiate between those two?  Both are failure conditions; we should continue to `return;` undefined.

test/streams.js
reqshark: stream default is a buffer, yes.

With node streams, we have a different way of asking the lib to return back a string. before everything node-nanomsg returned was a buffer

README.md
```diff
@@ -94,6 +94,48 @@ socket.connect('tcp://127.0.0.1:5555');
 
 *(Function, param: Function)*: Closes the socket. Any buffered inbound messages that were not yet received by the application will be discarded. The nanomsg library will try to deliver any outstanding outbound messages for the time specified by `linger`.
 
+## sending and receiving: writable and readable
+
+### socket.send(msg)
```
nickdesaulniers: Ha!  The most important method, and we didn't have it documented. :lollipop:

lib/index.js
reqshark: it reduces the highwatermark from 16K msgs to 16msgs, it's much more useful if we have a backpressure mechanism

README.md
```diff
@@ -94,6 +94,48 @@ socket.connect('tcp://127.0.0.1:5555');
...
+*(Function, param: String or Buffer)*: equivalent to `socket.send()` in [node.zeromq](https://github.com/JustinTulloss/zeromq.node).
```
nickdesaulniers: Let's remove the reference to zmq.

lib/index.js
reqshark: if we want to pass stuff at initialization like `utf8`, yep

README.md
```diff
@@ -94,6 +94,48 @@ socket.connect('tcp://127.0.0.1:5555');
...
+
+```js
+socket.send('hello from nanømsg!');
+```
+
+`send(msg)` is automatically invoked during `Writable` consumption of some other `Readable` stream. In that case a `Writable`'s `pipe()` method can be used to transmit across readable data sources. See [example for more detail](examples/writablepipe.js). The flow of data distributes to endpoint(s) determined by the particular socket type.
+
+```js
+var fs = require('fs');
+var source = fs.createReadStream(__dirname + 'filename.ext');
```
nickdesaulniers: Don't we need the path.join() stuff on this?  It works on unix's, but I don't think it's portable to windows?

README.md
```diff
@@ -94,6 +94,48 @@ socket.connect('tcp://127.0.0.1:5555');
...
+
+source.pipe(socket); //sends each chunk as a msg to socket's particular endpoint
+```
+
+### socket.on(data, callback)
```
nickdesaulniers: `'data'`

README.md
```diff
@@ -94,6 +94,48 @@ socket.connect('tcp://127.0.0.1:5555');
...
+*(Function, param order: String, Function)*: The `Readable` stream's `on()` function is an event listener registered with the `nanomsg c lib` that emits `'data'` events. To receive messages, pass the string `'data'` followed a callback containing a single data parameter.
```
nickdesaulniers: We can remove the info about the `nanomsg c lib`.

lib/index.js
reqshark: if you return nothing streams will blow up with an uncaught exception on the process. and especially so if the returned msg is ever `-1`. actually any number type for that matter. You can only give streams a string or a buffer basically.

Anyway in practice this only crops up for req/rep and surveyor/respondent scenarios IMO

README.md
reqshark: yup, snailzilla must have got to that

README.md
```diff
@@ -94,6 +94,48 @@ socket.connect('tcp://127.0.0.1:5555');
...
+
+```js
+socket.on('data', function (msg) {
+  console.log(String(msg)); //'hello from nanømsg!'
```
nickdesaulniers: Isn't it better to convert a Buffer to a String by being explicit about the encoding? `msg.toString('utf8')`

README.md
```diff
@@ -94,6 +94,48 @@ socket.connect('tcp://127.0.0.1:5555');
...
+});
+```
+
+The readable stream's `data` event is automatically invoked when piped to a `Writable` or `Transform` consumer stream. See [example for more detail](examples/transforms.js). Here `msgprocessor` is a transform you could pipe to a writable or the next transform:
+
+```js
+var through = require('through');
+
+var msgprocessor = through(function(msg){
+  console.log(String(msg)); //'hello from nanømsg'
```
nickdesaulniers: explicit encoding

examples/writablepipe.js
```diff
@@ -0,0 +1,15 @@
+var nano = require('..')
+var pub   = nano.socket('pub'),       push  = nano.socket('push');
+var sub   = nano.socket('sub'),       pull  = nano.socket('pull');
+pub.bind('tcp://127.0.0.1:3333');     push.bind('tcp://127.0.0.1:4444');
+sub.connect('tcp://127.0.0.1:3333');  pull.connect('tcp://127.0.0.1:4444');
+
+sub.setEncoding('utf8'); //sub socket will get utf8 strings instead of Buffers
```
nickdesaulniers: This should be documented in the readme. Something like `this can be used to always convert received messages to Strings of an explicit encoding` or something.

nickdesaulniers:
@asafyish, can you apply this patch and comment on the changes in memory usage?  I don't have a dataset large enough to test these changes on.  It might not lower memory spikes, I would just like to guard against them.

lib/index.js
```diff
@@ -193,11 +200,11 @@ Socket.prototype._startPollSend = function () {
...
 Socket.prototype._startPollReceive = function () {
-    if (!this._pollReceive) {
-        this._pollReceive = nn.PollReceiveSocket(this.binding, function (events) {
-            if (events) this._receive();
-        }.bind(this));
-    }
+  if (!this._pollReceive) {
+    this._pollReceive = nn.PollReceiveSocket(this.binding, function () {
+      this.push(this._receive());
+    }.bind(this));
```
nickdesaulniers: Something doesn't seem right here.  Is it because we no longer inherit from EventEmitter that we no longer get the param events?  We should make sure nn.PollReceiveSocket doesn't ever invoke the callback with an argument that we should be checking for truthiness, like we were previously.

lib/index.js
nickdesaulniers: Let's add a comment for future contributors :guitar: :penguin:

nickdesaulniers: Also, is that number configurable?  It seems arbitrary that `objectMode` translates to 16 msgs.  Is that 16 1B messages?  What about 1KB messages?

lib/index.js
nickdesaulniers: Then's let's consistently return `null` on failure.  Seems weird that sockets don't like `undefined`, but are fine with `null`.  It's a very subtle distinction to make between the two.

README.md
reqshark: aren't all JS strings `utf8`?  i knew about `.toString()`, i didnt know u could pass it `utf8`. 

Also i think `String(str)` looks better.

nickdesaulniers: Oh, the default is utf

nickdesaulniers: ok, let's make the README consistent, since the very first example uses toString

lib/index.js
reqshark: no look at our libuv code. the callback only returns when we can do a read on the socket.

the param is extraneous

lib/index.js
reqshark: ya it's definitely configurable

lib/index.js
reqshark: ya the binding is already doing the events truthiness check in C/C++

lib/index.js
nickdesaulniers: How (other than setting it to true)?

lib/index.js
nickdesaulniers: What line?

reqshark: the descriptor ready check actually happens here, https://github.com/nickdesaulniers/node-nanomsg/blob/master/src/node_nanomsg.cc#L257 that's what lets the callback happen, anyway it calls NanomsgReadable and in there technically the callback fires.

either way the callback never fires unless it's readable. i'm pretty sure.

reqshark: or here, rather, https://github.com/nickdesaulniers/node-nanomsg/blob/master/src/node_nanomsg.cc#L281 if you're looking at nn_recv, same thing, nn_send seems pretty isomorphic.

lib/index.js
reqshark: no you configure it from highWaterMark option.

https://iojs.org/api/stream.html#stream_new_stream_readable_options
or
https://nodejs.org/api/stream.html#stream_new_stream_readable_options

README.md
reqshark: ok ..if we must

lib/index.js
reqshark: not that subtle when you think about the fact that `null` is an object according to `typeof`, meanwhile `undefined` is its own type.

reqshark:
@nickdesaulniers, I just updated this per your review last month, it's ready for another round.

reqshark:
run the standard performance benchmarks to gage part of the trade-off involved in merging this.

```bash
$ git clone https://github.com/nickdesaulniers/node-nanomsg.git emitter
$ git clone https://github.com/nickdesaulniers/node-nanomsg.git streams
$ cd emitter && make && cd ../streams && git checkout streams && make

# run the standard perf tests or modify them to look at streams versus emitter in master
$ make bench
$ make perf
```

## here's what I got for `make bench`
#### `EventEmitter`:
```bash
node perf/local_lat.js tcp://127.0.0.1:5555 10 1000& node perf/remote_lat.js tcp://127.0.0.1:5555 10 1000 && wait
message size: 10 [B]
roundtrip count: 1000
average latency: 75.849 [us]
node perf/local_thr.js tcp://127.0.0.1:5556 10 100000& node perf/remote_thr.js tcp://127.0.0.1:5556 10 100000 && wait
message size: 10 [B]
message count: 100000
throughput: 138283 [msg/s]
throughput: 11.063 [Mb/s]
```
#### `Duplex`:
```bash
node perf/local_lat.js tcp://127.0.0.1:5555 10 10000& node perf/remote_lat.js tcp://127.0.0.1:5555 10 10000 && wait
message size: 10 [B]
roundtrip count: 10000
average latency: 80.347 [us]
node perf/local_thr.js tcp://127.0.0.1:5556 10 100000& node perf/remote_thr.js tcp://127.0.0.1:5556 10 100000 && wait
message size: 10 [B]
message count: 100000
throughput: 98127 [msg/s]
throughput: 7.85 [Mb/s]
```

### Performance trade-off discussion and the way forward

Obviously streams will have a big impact on `throughput`, but the latency cost, at least on my machine seems very low.

Overall, the cost is pretty significant, it seems to me, but I still propose that we merge and bump major.

People can use version 1 of this for the higher performance edition packing sockets inheriting from `EventEmitter`. And then people who want the Streams API interface could just use version 2. 

If we start introducing more cool new stuff after the V2 release, then definitely it behooves us to re-introduce `EventEmitter`, but merely as a non-default option passed to the socket during initialization. That would surely keep everything monolithic, and if we decide against such a thing there is always the possibility of maintaining an `EventEmitter` branch with separate minor version bumps in parallel. 

I would be willing to help implement socket option or help keep an `EventEmitter` version supported, whatever we decide if we move forward with V2. What do you think @nickdesaulniers? What do people think? BTW, I just realized this closes an [outstanding issue nanomsg.iojs](https://github.com/reqshark/nanomsg.iojs/issues/19).

#### A final note
`Duplex` will have no impact on the position of this module's performance compared to alternatives:
*<sub>1. Even after absorbing the streams perf hit, [this module will outperform zmq](https://twitter.com/rekshark/status/571078445444284416)</sub>*
*<sub>2. node-nanomsg will still provide [an order of magnitude higher performance](https://github.com/reqshark/perf) than node's net module</sub>*

nickdesaulniers:
I think the changes to lib/index.js can be greatly simplified, check out this patch:
```diff
diff --git a/lib/index.js b/lib/index.js
index 776cb32..12bc8f8 100644
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,6 +1,6 @@
 var nn = require('bindings')('node_nanomsg.node');
 
-var util = require('util');
+var emitStream = require('emit-stream');
 var EventEmitter = require('events').EventEmitter;
 
 /**
@@ -126,7 +126,7 @@ function Socket (type, opts) {
   }
 }
 
-util.inherits(Socket, EventEmitter);
+Socket.prototype = emitStream(new EventEmitter);
 
 Socket.prototype._protect = function (ret, unwind) {
     if(ret < 0) {
@@ -181,7 +181,7 @@ Socket.prototype._receive = function () {
     if (msg == -1) return;
 
     if (this.restore && typeof this.restore === 'function') msg = this.restore(msg);
-    this.emit('message', msg);
+    this.emit('data', msg);
 };
 
 Socket.prototype._startPollSend = function () {
@@ -340,7 +340,7 @@ function Device (sock1,sock2) {
     }
 }
 
-util.inherits(Device, EventEmitter);
+Device.prototype = emitStream(new EventEmitter);
 
 /**
  * sockopt API
diff --git a/package.json b/package.json
index aaeb274..1e0bf8c 100644
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "main": "lib/index.js",
   "dependencies": {
     "bindings": "1.2.1",
+    "emit-stream": "^0.1.2",
     "nan": "~1.8.x"
   },
   "devDependencies": {

```

reqshark:
@nickdesaulniers that will work! nice, :+1: let's just do that

when i apply that patch to the streams branch, i'm getting some test failures however, do you see that?

nickdesaulniers:
@reqshark that patch was written against master branch, not streams, so if it applies, well, that was pure luck.  I did not run the tests :cat: Let's run each test individually and see what's messed up.

nickdesaulniers:
Also, I did not check if we could pipe into a stream, only pipe out of.  We should check that.

reqshark:
would you like me to commit your patch to index.js here in this branch and go from there, since the tests are converted for streams?

nickdesaulniers:
yeah, in your local branch, checkout the version of lib/index.js, then apply my patch on top, and let's go from there.

reqshark:
i'll push it up so we can check it

nickdesaulniers:
Some things we can add in this PR: bump the major version number to 2.0.0, and add a CHANGELOG.md that briefly mentions the switch to 2.0.0 was because we remove message events in favor of data events.

reqshark:
ok good idea, if i can't get to it now will do later this evening

nickdesaulniers:
also, if we can't pipe into, we might be able to swap out the emitStream dependency with [this](https://github.com/pgte/duplex-emitter).

reqshark:
not sure this is streams2

reqshark:
I wish we could do it that easily, but we cant just use that handy converter if we want our sockets exposed as modern node streams that people can consume

PR #56, opened 2015-02-25
revised send operation: be equivocal on native buffer/strings

reqshark:
we should see some improved send performance:
  - eliminates the buffer check and conversion in javascript

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -132,17 +132,18 @@ Socket.prototype._protectArray = function (ret, unwind) {
     return ret[1];
 };
 
-Socket.prototype._send = function (buf, flags) {
+Socket.prototype._send = function (msg, flags) {
     if (this.closed) return;
 
     if(this.type == 'surveyor') {
         this._startPollReceive();
     }
 
-    if (this.transform && typeof this.transform === 'function') buf = this.transform(buf);
-    if (!Buffer.isBuffer(buf)) buf = new Buffer(buf);
-    return this._protect(nn.Send(this.binding, buf, flags), function () {
-        this.queue.unshift([buf, flags]);
+    if (this.transform && typeof this.transform === 'function') {
+      msg = this.transform(msg);
+    }
+    return this._protect(nn.Send(this.binding, msg, flags), function () {
+        this.queue.unshift([msg, flags]);
     });
 };
 
--- a/src/node_nanomsg.cc
+++ b/src/node_nanomsg.cc
@@ -199,21 +199,30 @@ NAN_METHOD(Shutdown) {
 NAN_METHOD(Send) {
   NanScope();
 
-  int s = args[0]->Uint32Value();
+  int flags = args[2].integer;
+  std::string *input;
 
-  if (!node::Buffer::HasInstance(args[1])) {
-    return NanThrowError("second Argument must be a Buffer.");
+  if (node::Buffer::HasInstance(args[1]->ToObject())) {
+
+    v8::Handle<v8::Object> object = args[1]->ToObject();
+
+    const char *data = node::Buffer::Data(object);
+    input = new std::string(data, node::Buffer::Length(object));
+
+  } else {
+
+    utf8 str (args[1]->ToString());
+
+    input = new std::string(*str);
   }
 
-  Local<Object> obj = args[1]->ToObject();
-  char *odata = node::Buffer::Data(obj);
-  size_t odata_len = node::Buffer::Length(obj);
-  int flags = args[2]->Uint32Value();
+  v8::Local<v8::Number> bytes;
 
-  // Invoke nanomsg function.
-  int ret = nn_send(s, odata, odata_len, flags);
+  bytes = NanNew<Number>(nn_send (S, input->c_str(), input->length(), flags));
 
-  NanReturnValue(NanNew<Number>(ret));
+  delete input;
+
+  ret(bytes);
 }
 
 NAN_METHOD(Recv) {


lib/index.js
```diff
@@ -133,17 +133,17 @@ Socket.prototype._protectArray = function (ret, unwind) {
...
 
-Socket.prototype._send = function (buf, flags) {
+Socket.prototype._send = function (msg, flags) {
     if (this.closed) return;
 
     if(this.type == 'surveyor') {
         this._startPollReceive();
     }
 
-    if (this.transform && typeof this.transform === 'function') buf = this.transform(buf);
```
nickdesaulniers: Does this break that test?

nickdesaulniers: Oops nvm

src/node_nanomsg.cc
```diff
@@ -205,23 +205,21 @@ NAN_METHOD(Shutdown) {
...
 
-    // Invoke nanomsg function.
-    int ret = nn_send(s, odata, odata_len, flags);
+  if (node::Buffer::HasInstance(args[1]->ToObject())) {
+    v8::Handle<v8::Object> object = args[1]->ToObject();
+    const char *data = node::Buffer::Data(object);
+    input = new std::string(data, node::Buffer::Length(object));
+  } else {
+    utf8 str (args[1]->ToString());
+    input = new std::string(*str);
```
nickdesaulniers: Where do we delete/free this string? It seems silly to convert the C string to a std::string then back again later using c_str().

src/node_nanomsg.cc
```diff
@@ -205,23 +205,21 @@ NAN_METHOD(Shutdown) {
...
-    char* odata = node::Buffer::Data(obj);
-    size_t odata_len = node::Buffer::Length(obj);
-    int flags = args[2]->Uint32Value();
+  int flags = args[2].integer;
+  std::string *input;
 
-    // Invoke nanomsg function.
-    int ret = nn_send(s, odata, odata_len, flags);
+  if (node::Buffer::HasInstance(args[1]->ToObject())) {
+    v8::Handle<v8::Object> object = args[1]->ToObject();
```
nickdesaulniers: How expensive it ToObject()? Do we need to call it twice or can we move the creating of object to be outside the conditional?

nickdesaulniers:
So this allows you to call send with either a JS string or a buffer? or we already had that and this improves perf?

src/node_nanomsg.cc
reqshark: not sure, adding the perf tests now.. we need to benchmark it.

reqshark:
this PR removes: 
```diff
-    if(!node::Buffer::HasInstance(args[1])) {		
-        return NanThrowError("second Argument must be a Buffer.");		
-    }
```

until now you could only pass buffers to the binding. `str` or `buf` conversion was handled in `index.js`

this change lets us pass either a JS string or a Buffer down into the binding from `index.js`

src/node_nanomsg.cc
reqshark: eh true, for all i know that `str` could be billowing out onto the heap.

I must have wanted `nn_send()` to go down outside the condition i guess

but now that you mention it we need to fix those two things.. let's just confirm the performance first, I don't want to make this change if it's faster to send the binding a buffer and use node javascript to do the conversion (which i kinda doubt, but who knows)

so i'll run some benchmarks on small and large message envelopes later tonight and confirm then

reqshark:
so I ran some perf tests and found this PR has zero discernible impact on latency and throughput when it comes to buffers. 

But there seems to be a slight advantage when it comes to sending strings. :+1: 

Actually that's not the reason we need to replace our `send()` method implementation asap. It's actually not as safe and reliable as we need. 

When interfacing over remote protocols with other native implementations that do not use `node::buffer` class there seems to be a random possibility of msg corruption. It may be extremely difficult to detect at times and at other times it could occur with high frequency.  

We've dealt with this on the `nn_recv()` side at reqshark/nanomsg.ios by debugging and rewriting the memory allocation mechanism inside the iOS port of libnanomsg, see e.g. https://github.com/reqshark/nanomsg.ios/pull/5 and https://github.com/reqshark/nanomsg.ios/pull/3 and https://github.com/reqshark/nanomsg.ios/commit/043cedfef32882467e6404e73e0715c5fa2f7201 

This work has proven to give us a reliable and stable version of libnanomsg that's been battle tested against the common programmer error of sending a C string over the wire without null terminating.

You might think node::buffer handles data safely by null terminating a `char` type especially as it uses a known length to allocate heap for such a pointer. 

Granted, V8 engineers have often made deliberate choices not to null terminate strings and probably the best example is when they decided against it for utf8.

Both node::buffer class and node::smalloc seem to handle char pointers without null terminating them, so we will need to pick up the slack there and do it ourselves.

Another reason to do that is because allocations during `nn_recv()` I/O operations are performed using `malloc`, the merits of which have been debated about at length on [nanomsg PR#363](https://github.com/nanomsg/nanomsg/pull/363) and so the library's position about that now is very clear: it is not the library's responsibility to ensure these blobs are safely dispatched before handing off to `nn_send()`, no no that, my friends, falls square on us.

The C++ std::string class does a convenient C string conversion with a safe null termination as well as the added benefit of freeing the allocated memory when its calling context returns execution control back to the application. Let's just do that and cut out the extra buffer conversion code from our JS file, passing strings or buffers without a care for type directly into our binding from now on! 

ok I'm rebasing this..

src/node_nanomsg.cc
```diff
@@ -199,21 +199,24 @@ NAN_METHOD(Shutdown) {
...
-  Local<Object> obj = args[1]->ToObject();
-  char *odata = node::Buffer::Data(obj);
-  size_t odata_len = node::Buffer::Length(obj);
-  int flags = args[2]->Uint32Value();
+    v8::Handle<v8::Object> object = args[1]->ToObject();
 
-  // Invoke nanomsg function.
-  int ret = nn_send(s, odata, odata_len, flags);
+    const char *data = node::Buffer::Data(object);
+    input = new std::string(data, node::Buffer::Length(object));
```
nickdesaulniers: there's no point in putting data (char*) into input (std::string) just to put it back in to a (char*) in the ret macro.  Just move ret into the branches of the conditional, and for this branch don't waste time constructing a std::string.

reqshark: giving a char* to std::string lets us null terminate the buffer and free that allocation.

```c
const char *data = node::Buffer::Data(object)
```
`data` is not a null terminated string, that's the problem. and we still have to free it later.

nickdesaulniers: If a node::buffer::Data object can tell you its length, then you can explicitly set the last char (len - 1) to `'\0'`.  No need to construct a std::string just to zero out the last byte.  Instead, what you should do is create a `char []`, whose length is node::Buffer::Data's length + 1, memcpy the data, then explicitly set the last byte to `'\0'`.  This will overflow the stack for messages larger than 5 MB on most OS', so maybe a length check and a warning or heap allocation.

reqshark: that makes sense to me, but keep in mind @asafyish is talking about sending msgs that are hundreds of mbs https://github.com/nickdesaulniers/node-nanomsg/pull/53#issuecomment-84999155

nickdesaulniers: ok, but this is still a memory leak.  The return value from NanNew<Number> should be stored in a temporary, then input deleted, then return the temporary.

reqshark: Doesn't std::string free the allocation? input gets deleted by the C++ class. And NanNew creates a temporary reference that gets recycled when nanscope is closed, so that gets taken care of.

nickdesaulniers: NanNew will hand a reference to a Number, the int returned from nn_send, not the input argument passed to nn_send.  std::string will deallocate memory, when it goes out of scope when stack allocated.  When heap allocated, you have to call delete explicitly on it.

reqshark: Ok actually that sounds right, when I get off this train I'll update the diff

nickdesaulniers:
nice!  waiting on Travis.  I just triple checked this in C++ with valgrind:
```c++
// clang++ -g temp.cpp && valgrind --leak-check=full ./a.out
#include <string>
int main () {
  std::string a = "Hello";
  std::string* b = new std::string("leak me");
  std::string* c = new std::string("free me");
  delete c;
}
```
prints:
```
...
==66207== HEAP SUMMARY:
==66207==     in use at exit: 58,843 bytes in 352 blocks
==66207==   total heap usage: 538 allocs, 186 frees, 64,019 bytes allocated
==66207==
==66207== 40 (8 direct, 32 indirect) bytes in 1 blocks are definitely lost in loss record 27 of 82
==66207==    at 0x6B1B: malloc (in /usr/local/Cellar/valgrind/3.10.1/lib/valgrind/vgpreload_memcheck-amd64-darwin.so)
==66207==    by 0x4E346: operator new(unsigned long) (in /usr/lib/libstdc++.6.0.9.dylib)
==66207==    by 0x100000BDF: main (temp.cpp:4)
...
```

where temp.cpp:4 is `std::string* b = new std::string("leak me");`

nickdesaulniers:
sizzurp!!! baaaaaaaaah :sheep:

nickdesaulniers:
we should do a minor version bump, since now you can send strings!

reqshark:
ya bump version, we stay sippin!  :+1:

nickdesaulniers:
so, finally got around to reading this thread.  I wonder if we accidentally just introduced unnecessary null termination:

https://github.com/nanomsg/nanomsg/pull/363#issuecomment-76807351

I'm quite sure we did.  We can use 

```c
for (int i = offset; i < len; ++i) {
  printf("%hhx\n", buffer[i]);
}
```

to view the bytes of a non null terminated char[]. `%s` only works with null terminated strings.

reqshark:
without null termination, nn_recv operations in other native implementations may contain corrupt byte data at the end of the message.

nickdesaulniers:
That sounds like incorrect conversion on the receiving side of raw non-null-terminated byte array to native string representation.

nickdesaulniers:
> reqshark: so I have a question, what's your definition of a byte array?

a contiguous sequence of bytes in memory that may or may not contain a null terminating byte, `\0`.  In C:

```c
char a [] = { 0, 1, 2 }; // no null termination, byte array, not a valid C string
char b [] = { 0, 1, 2, '\0' }; // null terminated, byte array, valid C string
char c [] = "012"; // null terminated, byte array, valid C string
// there is no mirror for creating a non-null terminated byte array using the literal "" notation
```

like a square can be a rectangle, but a rectangle might not be a square; in C, a string can be a byte array but a byte array might not be a string.

reqshark:
that's correct, https://github.com/reqshark/nanomsg.ios/commit/20b8a79994078d70956b5bf43fc6725ad2d14953 introduced a regression that I was working against, which regardless of byte arrays or c style strings, lead me to this belief.

reqshark:
sorry for the red herring, there, I think
> we accidentally just introduced unnecessary null termination

nickdesaulniers:
https://github.com/nickdesaulniers/node-nanomsg/issues/74

PR #59, opened 2015-02-27
Perf tests translated from nanomsg

comick:
See https://github.com/nanomsg/nanomsg/tree/master/perf as a reference.

Recently added perf scripts, seem to be translated directly from node zmq bindings.
I believe we should target nanomsg for performances comparison:

- same output
- same socket types
- same time units make more sense to me.

A part from obvious translation, the use of setTimeout is replaced with ```flush```.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,12 +13,12 @@ clean:
 	rm -fr build && rm -rf node_modules
 
 perf:
-	node perf/local_lat.js tcp://127.0.0.1:5555 1 100000& node perf/remote_lat.js tcp://127.0.0.1:5555 1 100000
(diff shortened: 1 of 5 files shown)


nickdesaulniers:
Nice! @reqshark can you review?

reqshark:
in about 45 mins when I get off this train

reqshark:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

reqshark:
merged fb0b0571c6e54a21cab1b0c722e1184b1f0c5aa1! thank you!

comick:
Cool!

PR #37, opened 2014-09-13
Fix PUB/SUB topics support

drr00t:
Hi, PR to fix #31 issue, we need expose internal constants in a better way, i don't like this.

```javascript
var nn = nano._bindings;
sub.setsockopt(nn.NN_SUB, nn.NN_SUB_SUBSCRIBE, 'my.topic');
```

thank you for bind Nick.

Issue #31, opened 2014-05-21
Pub/Sub on a particular topic?


pselden:
I'm having trouble subscribing/publishing to a particular socket. Whatever I do, my subscriber ends up receiving all messages.

Is this not implemented?

```javascript
// subscriber.js
var nano = require('nanomsg');

var sub = nano.socket('sub');
var nn = nano._bindings;

var addr = 'tcp://127.0.0.1:7789'
sub.setsockopt(nn.NN_SUB, nn.NN_SUB_SUBSCRIBE, 'my.topic');
sub.connect(addr);

sub.on('message', function (buf) {
    console.log(buf.toString());
});

// publisher.js
var nano = require('nanomsg');
var pub = nano.socket('pub');
var addr = 'tcp://127.0.0.1:7789'
pub.bind(addr);

setInterval(function () {
    pub.send("my.topic This is a topic");
    pub.send("my.dog My dog's name is Spot");
}, 1000);
```

asafyish:
+1

nickdesaulniers:
I think this might actually be broken upstream.  See https://github.com/nickdesaulniers/node-nanomsg/issues/51#issuecomment-77778381

PR #61, opened 2015-02-27
optval pointer fix

reqshark:
@nickdesaulniers, following up a discussion we had about getsockopt args in #46:

> The third arg, optval, should be a void*. Passing a int [64] is probably not what you meant to do?

I'm pretty sure `NAN` or maybe node V8 is partial to `char*` over `void*`... though i guess it depends on the situtation/function we're talking about

> This does work, it's just using 63 * sizeof(int) bytes more of the stack than it needs to. It works because in C arrays decay into pointers when passed to functions. So you're allocating space for 64 ints on the stack, then passing the address of the first element, which gets filled. Since it's an array of ints, only the first element is ever populated.

> If you think about the C ABI, it's not possible to represent 64 ints in a single register on today's processors. That's why passing an array decays into a pointer. A kind of syntactic sugar. nn_getsockopt takes a void*, which works since optval decays into an int*.

> A few lines down, there is no need to access optval[0] since we don't need optval to be an array of ints, just an int.

agreed.

### DIFF
--- a/src/node_nanomsg.cc
+++ b/src/node_nanomsg.cc
@@ -148,17 +148,13 @@ NAN_METHOD(Setopt) {
 NAN_METHOD(Getopt) {
   NanScope();
 
-  int optval[1];
-  int option = args[2].integer;
-  size_t optsize = sizeof(optval);
+  size_t optsize;
+  int optval, option = args[2].integer;
+  optsize = sizeof (optval);
 
-  //check if the function succeeds
-  if(nn_getsockopt(S, args[1].integer, option, optval, &optsize) == 0){
-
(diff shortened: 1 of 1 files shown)


nickdesaulniers:
I'm sorry, this is not right.  I will fix this when I get a chance, but the gist is we have to do something special in the case of pub sub.  Se also #31 and #37.

nickdesaulniers:
See also how get and set socket option is used within nanomsg/nanomsg.  This gist is for some options we need one single integer (no arrays) (for pretty much every option) or a character array (for pub sub).

reqshark:
oh ok, is there string use other than subscription keys/channels?

nickdesaulniers:
I don't think so, but I need to re analyze nanomsg/nanomsg to triple check.

PR #62, opened 2015-02-28
perf node's net and udp modules for comparison

reqshark:
sometimes the situation calls for a different protocol, for example: UDP sockets establish the real limits of throughput

benchmark the options.

`thr` and `lat` tell us a lot depending on where and how they are run across the network (espeically different message sizes).

to compare node's TCP socket api, the `net` module run:

```bash
$ make net
```

for the standard datagram sockets shipped with io.js and node run:
```bash
$ make udp
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,11 @@
-.PHONY: clean check test perf bench full
+.PHONY: clean check test perf bench net udp full
 
 ALL:
+	git submodule update --init
 	npm i
 
 check:
-	npm t
+	npm i && npm t
 
 test:
 	npm t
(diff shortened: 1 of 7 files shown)


nickdesaulniers:
nice.  @comick can you review?

comick:
In my opinion we should not benchmark against node udp, tcp, whatever.
The point of using nanomsg is its scalable protocols and other useful features which are not in tcp, udp. nanosg will probably add udp support at some point.

reqshark:
node-nanomsg offers a latency improvement by at least an order of magnitude over node's TCP socket.

As for UDP my thinking was that messages would express the limit of the scripting language's built in latencies, however, this mechanism only demonstrates built-in maximum throughput.

on the other hand, tap a kernel UDP api in C: this actually expresses the system latency top-ends.

these tests illustrate major benefits associated with using node-nanomsg for msg latency, and when nanomsg is capable of a UDP API, I believe this will compare unreliable throughput.

these may not belong in the repo, but as informative tests, I remember when I wrote them, I had the feeling, where do i put this? while these tests should be run and observed, @comick, i have to agree. 

@nickdesaulniers what do you think?

comick:
Those scripts are certainly useful in making a choice whether using one library or another.
The common use case I see is that people know about nanomsg, made a choice against zmq (or any other mq), then they look-up for node binding.
In my opinion our parformance target should only be nanomsg reference C implementation.

reqshark:
> node-nanomsg performance target should only be the nanomsg reference C implementation.

+1

PR #65, opened 2015-03-16
pubsub: start work on the channel api

### LINKED ISSUE #31: Pub/Sub on a particular topic?

reqshark:
#### UPDATE (3/21/2015): API is compatible and extended:
*  the default sub socket will remain as it was: initialized with auto-subscription to *all-channels*.
*  `socket.chan()` calls internal check for the *all-channels* listener, removing it if necessary 

New methods:

### socket.rmchan(*comma separated*)

the new API lets us do stuff like:
```js
var sub = nano.socket('sub')
sub.chan(['AAPL', 'MSFT', 'whatever'])

sub.on('message', function(msg){
  // only get msgs prefixed by those registered by .chan()
  console.log(String(msg))
})
``` 

~~Also I'd like to take advantage of `NN_SUB_UNSUBSCRIBE` like~~:
UPDATE (3/21/2015): We now take advantage of `NN_SUB_UNSUBSCRIBE`:
```js
sub.rmchan('MSFT', 'whatever')

sub.on('message', function(msg){

  // only get msgs prefixed 'AAPL' (following from snippet above)
  console.log(String(msg))
})
```

~~I haven't updated tests, nor added a handler for `chans` given to `opts` object at socket initialization~~:

```js
var nano = require('nanomsg')
var sub = nano.socket('sub', {
  chan: ['AAPL', 'MSFT', 'whatever']
})
```

~~I will follow up and commit those two things next if this API change meets with consensus/approval.~~

With that, let's open the forum to all opinions/arguments in favor or against this change.


lib/index.js
```diff
@@ -315,6 +316,26 @@ Socket.prototype.maxreconn  = opt('maxreconn');
 Socket.prototype.sndprio    = opt('sndprio');
 Socket.prototype.rcvprio    = opt('rcvprio');
 
+/* chan and rmchan sockopt methods. only relevant for subscription sockets */
+Socket.prototype.chan       = function(){
+  var args = arguments;
+  var channels = args.length;
+  while(--channels > -1) {
```
nickdesaulniers: `while (--channels) {` is a common JS idiom, since `0` is false-y.

lib/index.js
```diff
@@ -315,6 +316,26 @@ Socket.prototype.maxreconn  = opt('maxreconn');
 Socket.prototype.sndprio    = opt('sndprio');
 Socket.prototype.rcvprio    = opt('rcvprio');
 
+/* chan and rmchan sockopt methods. only relevant for subscription sockets */
+Socket.prototype.chan       = function(){
+  var args = arguments;
```
nickdesaulniers: Let's get rid of this additional variable.

lib/index.js
```diff
@@ -315,6 +316,26 @@ Socket.prototype.maxreconn  = opt('maxreconn');
...
+  var channels = args.length;
+  while(--channels > -1) {
+    if(nn.Setopt(this.binding, nn.NN_SUB, nn.NN_SUB_SUBSCRIBE, args[channels]) == 0)
+      this.channels[args[channels]] = 'active';
```
nickdesaulniers: let's just assign this as a boolean instead of the string 'active'.  Also, please use `{}` with single statement if statements.

lib/index.js
```diff
@@ -315,6 +316,26 @@ Socket.prototype.maxreconn  = opt('maxreconn');
...
+  };
+}
+
+Socket.prototype.rmchan   = function(channel){
+  var args = arguments;
+  var channels = args.length;
+  while(--channels > -1) {
+    if(nn.Setopt(this.binding, nn.NN_SUB, nn.NN_SUB_UNSUBSCRIBE, args[channels]) == 0)
```
nickdesaulniers: We should check that the channel is even set before calling into C++ unsubscribe.

src/node_nanomsg.cc
```diff
@@ -37,11 +37,11 @@ NAN_METHOD(Socket) {
...
-    if( (ret >= 0) && (protocol == NN_SUB)) {
-        if (nn_setsockopt(ret, NN_SUB, NN_SUB_SUBSCRIBE, "", 0) != 0) {
-            return NanThrowError("Could not set subscribe option.");
-        }
-    }
+    //if( (ret >= 0) && (protocol == NN_SUB)) {
+        //if (nn_setsockopt(ret, NN_SUB, NN_SUB_SUBSCRIBE, "", 0) != 0) {
+            //return NanThrowError("Could not set subscribe option.");
+        //}
+    //}
```
nickdesaulniers: Let's just remove this.

src/node_nanomsg.cc
```diff
@@ -136,7 +136,7 @@ NAN_METHOD(Setopt) {
   int level = args[1].integer;
   int option = args[2].integer;
 
-  if(option == NN_SOCKET_NAME){
+  if(option == NN_SOCKET_NAME || option == NN_SUB_SUBSCRIBE){
```
nickdesaulniers: Hmm, so this isn't quite complete yet, right?

lib/index.js
reqshark: Oh I did that since otherwise writing `arguments` makes this line quite long for readability, are you sure?
```js
if(nn.Setopt(this.binding, nn.NN_SUB, nn.NN_SUB_SUBSCRIBE, args[channels]) == 0)
```

lib/index.js
reqshark: oh cool, but i think i need to include 0 here

lib/index.js
reqshark: good call

src/node_nanomsg.cc
reqshark: ... good catch

lib/index.js
nickdesaulniers: Not if you subtract 1 from your array index, or use post decrement.
```
while (--i > -1) { arr[i]; }
```
compared to:
```
while (i--) { arr[i] }
```

reqshark: sweet!

reqshark:
here's a detailed summary of the last commit and what we're doing for sub sockets:

updated test files with subscription sockets:
* `test/inproc.js`
* `test/ipc.js`
* `test/sockoptapi.js`
* `test/tcp.js`
* `test/transform.js`

added test file demonstrating some of the new channel api:
* `test/chan.js`

removed test files:
* `test/getsockopt.js`
* `test/setsockopt.js`

these were duplicates since documented `sockopts` met test coverage plus some additional.

in `src/node_nanomsg.cc`, removed code:
*  `NAN_METHOD(Setsockopt)`
*  `NAN_METHOD(Getsockopt)`

and simplified existing methods by dropping support for `NN_SOCKET_NAME`:
* `NAN_METHOD(Setopt)`
* `NAN_METHOD(Getopt)`

added a new method to handle subscribe/unsubscribe:
* `NAN_METHOD(Chan)`

we're keeping socket option code restrained by separating the subscription handling concern out into its own small function. the rule of modularity asks us to do one thing and do it well.

Although socket channel registration is performed by `nn_setsockopt()`, @nickdesaulniers and I discussed and found shifting the operation into `NAN_METHOD(Chan)` reduces C pointer type complexity ordinarily arising from an encapsulated `Setsockopt()` method. This alternative also provides a concise and complete interface underlying API calls to `socket.chan()` and `socket.rmchan()`.

with fewer tests there are exactly 800 nyan-cat, all green :palm_tree:

lib/index.js
```diff
@@ -96,16 +96,38 @@ function Socket (type, opts) {
...
-    for(var sokopt in sol){
-        if(opts.hasOwnProperty(sokopt)) this[sokopt](opts[sokopt]);
+  /* subscription filters */
+  this.channels = {};
+
+  /* subscription handling at initialization */
+  if (opts.hasOwnProperty('chan')) {
+    if (typeof opts.chan == 'string') {
+      opts.chan.split(',').forEach(subscriber.bind(this))
+    } else {
```
nickdesaulniers: This `else` branch should be an `else if (Array.isArray(opts.chan)) {`

lib/index.js
```diff
@@ -96,16 +96,38 @@ function Socket (type, opts) {
...
+      opts.chan.forEach(subscriber.bind(this))
     }
```
nickdesaulniers: In both cases of the conditional, the code would be more robust if we first called `.trim()` on what was being passed to subscriber.  In the case of splitting, we should filter out empty elements, after trimming.  If you map trim across all elements, then filter on the first branch, that would be nice.

lib/index.js
```diff
@@ -96,16 +96,38 @@ function Socket (type, opts) {
...
+  this.queue = [];
 
-    for(var sokopt in sol){
-        if(opts.hasOwnProperty(sokopt)) this[sokopt](opts[sokopt]);
+  /* subscription filters */
+  this.channels = {};
+
+  /* subscription handling at initialization */
+  if (opts.hasOwnProperty('chan')) {
+    if (typeof opts.chan == 'string') {
```
nickdesaulniers: Let's use strict equality everywhere, always.

lib/index.js
```diff
@@ -96,16 +96,38 @@ function Socket (type, opts) {
...
     }
+  }
 
-    if(this.af_domain == nn.AF_SP) {
-        if (this.receiver) this._startPollReceive();
+  function subscriber(chan) {
+    if (nn.Chan(this.binding, nn.NN_SUB_SUBSCRIBE, chan) != -1) {
+      this.channels[chan] = true;
+    } else {
+      console.log('sub err')
```
nickdesaulniers: Let's log this to stderr with console.error.

lib/index.js
```diff
@@ -372,6 +394,28 @@ function opt (option) {
   }
 };
 
+/* chan and rmchan sockopt methods. only relevant for subscription sockets */
+Socket.prototype.chan       = function(){
```
nickdesaulniers: One of the issues I have with the current API is that during socket construction, you can pass either a comma separated string, or an array.  After construction, you can call the chan method with only an array.

Since you might want to filter on chans that contain a comma, splitting on the comma in the constructor seems kind of brittle.  Rather than know about that edge case, why don't we eliminate it and require library consumers to have to always pass an array of strings?

lib/index.js
```diff
@@ -372,6 +394,28 @@ function opt (option) {
...
+  var args = arguments;
+  var channels = args.length;
+  while(channels--) {
```
nickdesaulniers: Would prefer that variables used for the sole purpose of iteration be named `i`, `j`, or `k`.

lib/index.js
```diff
@@ -372,6 +394,28 @@ function opt (option) {
...
+  };
+}
+
+Socket.prototype.rmchan     = function(){
+  var args = arguments;
+  var channels = args.length;
+  while(channels--) {
+    if (this.channels[args[channels]] &&
+      nn.Chan(this.binding, nn.NN_SUB_UNSUBSCRIBE, args[channels]) != -1) {
+        this.channels[args[channels]] = false;
```
nickdesaulniers: Maybe we should be deleting the key from the obj, rather than setting it false?  Eh, I'm on the fence.  Thoughts?

test/chan.js
```diff
@@ -0,0 +1,75 @@
...
+
+    var msg = String(buf);
+
+    //increment msgqty when a msg is received (happening about every 100ms)
+    //we'll call finish() on the 5th msg received (case 4, see line 68)
+    switch(++msgqty){
+      case 0: return t.equal(msg, msgs['m' + msgqty]); //msgs.m0, 'hello world'
+      case 1: return t.equal(msg, msgs['m' + msgqty]); //msgs.m1, 'hello world'
+      case 2: return t.equal(msg, msgs['m' + msgqty]); //msgs.m2, 'hello world'
+      case 3: return t.equal(msg, msgs['m' + msgqty]); //msgs.m3, 'bar world'
```
nickdesaulniers: just make msgs an array, then here you can `return t.equal(msg, msgs[msgqty])`

test/chan.js
```diff
@@ -0,0 +1,75 @@
...
+  sub.connect('inproc://filter');
+
+  //the main object we'll use to establish channel registration and removal
+  var msgs = {
+    m0:'hello world',
+    m1:'hello world',
+    m2:'hello world',
+    m3:'bar world',
+    m4:'foo world'
+  }
```
nickdesaulniers: we should add a message that is not a chan we're subscribed too and explicitly fail if that message is encountered.

test/chan.js
nickdesaulniers: ah nevermind, I see that's what removeHello is doing

test/inproc.js
```diff
@@ -10,7 +10,7 @@ test('inproc socket pub sub', function (t) {
     t.plan(1);
 
     var pub = nano.socket('pub');
-    var sub = nano.socket('sub');
+    var sub = nano.socket('sub', {chan:''});
```
nickdesaulniers: While it's nice to see you're being explicit about not subscribing to any chans, this makes it look like the second arg to nano.socket is required.  Let's remove the explicitly empty options argument, that way users reading the tests for usage don't think that they are required to construct a meaningless object.

test/mksymbols.rb
```diff
@@ -42,4 +42,3 @@
 #{o}
 ];
 FOO
-
```
nickdesaulniers: Let's try to keep new lines at the end of files, see: http://stackoverflow.com/questions/729692/why-should-files-end-with-a-newline

nickdesaulniers:
Looks good!  A few minor nits, the biggest being that I think we should always expect an array of strings.

This will require a major version bump, since we're removing the getsockopt and setsockopt api, and requiring you to go through the getters/setters.

Patches that remove more code than they add :+1: :+1: :+1: Thanks @reqshark !

src/node_nanomsg.cc
```diff
@@ -57,86 +51,14 @@ NAN_METHOD(Close) {
...
 
-  if (option == NN_SOCKET_NAME) {
-    utf8 str(args[3]);
-    ret(NanNew<Number>(nn_setsockopt(S, level, option, *str, str.length())));
-  } else {
-    int optval = args[3].integer;
-    ret(NanNew<Number>(
-        nn_setsockopt(S, level, option, &optval, sizeof(optval))));
-  }
+  ret(NanNew<Number>(nn_setsockopt(S, level, option, &optval, sizeof(optval))));
```
nickdesaulniers: make sure you run `clang-format -style=Mozilla -i <file>` on this!

nickdesaulniers:
Please edit your commit messages to mention that they `Fixes #67, Fixes #31`.  Then, GH will auto close those issues when the patch is merged.

lib/index.js
reqshark: OK but if that thing is neither String nor Array then I want that forEach to throw the process

lib/index.js
reqshark: oh nice! started to think about that too after I saw how that string looked

lib/index.js
reqshark: word

lib/index.js
reqshark: I was thinking maybe we should throw here actually.

One thing that will come up in another PR (forthcoming related to `nn_shutdown()`), is that we'll need to decide about dynamic re-subscription, according to [folks on the mailing list](http://www.freelists.org/post/nanomsg/Question-on-pubsub-device,5) it may not be possible, and backing that idea up, I'm seeing some unverified issues with latent calls after `socket.rmchan()` on a specific topic.

lib/index.js
reqshark: > After construction, you can call the chan method with only an array.

well `arguments` is certainly array-like. after socket construction `chan()` accepts variable arity, so it's just comma separated parameters

lib/index.js
reqshark: I personally would understand the code, and it would look cleaner to me with less of it, but what about someone new to the project? 

Is short `var` more legible than a descriptive `var`?

As I recall ZeroMQ is somewhat opinionated about that, of course that's not necessarily the style of libnanomsg, whose naming conventions are more aligned with POSIX standards than probably anything else.

lib/index.js
reqshark: definitely, good idea, I want to use the `delete` operator

test/chan.js
reqshark: for the test ya i agree, that was my purpose for writing the `removeHello` function

test/inproc.js
reqshark: from the beginning with these experiments [my philosophy was to set options early](https://github.com/reqshark/nanomsg.iojs/tree/a49cacc15bb8c6ca195b50d2ea780f4f92bec3fd#philosophy-setsockopts-early).

i think it's better practice than calling `.chan()` later, but sure, we can do that

test/mksymbols.rb
reqshark: weird i didn't meant to do that.

src/node_nanomsg.cc
reqshark: ok will do before i update my changes. 

btw, i'll try to preserve the commit history, the reason it's lost before this point is that merge conflict yesterday..

reqshark:
> Please edit your commit messages to mention that they Fixes #67, Fixes #31. Then, GH will auto close those issues when the patch is merged.

yep before merging, I just figured if i needed to rebase or whatever i didn't want to keep re-referencing phantom commits

lib/index.js
nickdesaulniers: Then you can spread them out:

```js
chan.apply(this, Array.prototype.slice.call(arguments));
```

test/inproc.js
```diff
@@ -15,6 +15,8 @@ test('inproc socket pub sub', function (t) {
     var addr = 'inproc://pubsub';
     var msg = 'hello world';
 
+    sub.chan('');
```
nickdesaulniers: Ah, I meant can we remove all references to the `''` chan from tests that don't test filtering on a sub socket?  Will the sub socket fail to work, since I think we always have to at least subscribe to the empty chan?  If so, since we can have multiple subscription, then we should call `this.chan('')` during socket construction if it's a sub socket.  Otherwise, it's silly to force users to have to explicitly subscribe to the empty channel just to use sub sockets.

test/mksymbols.rb
nickdesaulniers: bump

test/chan.js
nickdesaulniers: bump

lib/index.js
```diff
@@ -96,16 +106,40 @@ function Socket (type, opts) {
...
-    if(this.af_domain == nn.AF_SP) {
-        if (this.receiver) this._startPollReceive();
+  /* subscription handling at initialization */
+  if (opts.hasOwnProperty('chan')) {
+    if (typeof opts.chan === 'string') {
+      opts.chan.split(',').map(function(i){
+        return i.trim();
+      }).forEach(subscriber.bind(this))
+    } else if (Array.isArray(opts.chan)) {
+      opts.chan.forEach(subscriber.bind(this))
```
nickdesaulniers: in both cases, we should map trim, then filter empty elements.  This prevents `chan: ['' ' a', 'b ', ' c ']` and `chan: ", a,b , c "`.

lib/index.js
```diff
@@ -4,6 +4,16 @@ var util = require('util');
...
 
 /**
+ * routine errors
+ */
+var er = {
+  op              : 'nanomsg operation failed: ',
+  param           : 'parameter type err: ',
+  chan            : 'setting channel subscription string: ',
+  rmchan          : 'removing channel subscription string: ',
+}
```
nickdesaulniers: There's really no need for this, just put the strings in place.  DRY has its place, but not with error message substrings.

lib/index.js
```diff
@@ -96,16 +106,40 @@ function Socket (type, opts) {
...
+      }).forEach(subscriber.bind(this))
+    } else if (Array.isArray(opts.chan)) {
+      opts.chan.forEach(subscriber.bind(this))
+    } else {
+      throw new Error( er.param + 'sub channel uses String or Array\n');
     }
+  }
+
+  function subscriber(chan) {
+    if (nn.Chan(this.binding, nn.NN_SUB_SUBSCRIBE, chan) !== -1)
```
nickdesaulniers: curly braces

lib/index.js
nickdesaulniers: Further, if we only take an array of strings, then there's no ambiguity about whitespace, and we don't need to filter or map the input.

test/inproc.js
reqshark: > can we remove all references to the '' chan from tests that don't test filtering on a sub socket?

that would be useful. maybe if we run a setTimeout or Interval after sub socket is created that checks for the existence of any length on `this.channel`. 

If we did something like that it would make more sense to only accept arrays for topics, since we could just check length on that property.

well if you have a better implementation idea in mind I'm all ears, but the main thing is that once we've subscribed to `''`, there's no point in registering subsequent subscription topics, since these plus all others would get filtered into the `message` event callback

lib/index.js
reqshark: ya the need is all these additional calls to `new Error('...` are longer than 81 chars per line.

I'd rather one-line new Errors, but if you insist.

well, hey I had some other repeated errs in mind too, it's not just those.

eh, I can take it out I guess, but you'd really rather see calls to `new Error` broken into multiple lines, or worse: trailing off beyond 80 chars to the right of the screen?

test/inproc.js
nickdesaulniers: > that would be useful. maybe if we run a setTimeout or Interval after sub socket is created that checks for the existence of any length on this.channel.

Then you run into the polling problem.  What duration do you use? If too long then you miss messages, if too short then you are subject to v8/machine perf.  Too brittle.

So, if you don't subscribe to any channels explicitly, sub sockets don't work (can we confirm)?

> once we've subscribed to '', there's no point in registering subsequent subscription topics

Can we make `''` the default, then the first time someone calls `chan`, can we `rmchan('')`?

reqshark: > Can we make '' the default, then the first time someone calls chan, can we rmchan('')?

excellent idea.

test/mksymbols.rb
reqshark: github atom editor doesn't let me fix this file. that's what i've been using lately.

i think there is a line when you pull it down. whatever i use vim to fix that

lib/index.js
nickdesaulniers: We'll let the formatter figure it out, when it lands.  I don't have my heart set on 80 char line limits.

test/mksymbols.rb
nickdesaulniers: Use the git command line and remove it from the patch.  http://stackoverflow.com/questions/5020305/how-can-i-remove-a-change-in-1-file-in-1-commit

lib/index.js
reqshark: ok i was holding back since `nano.socket('sub', { chan: [''] })` is unacceptable... 
... as far as style goes.

but since you came up with that idea of `''` as the defualt where we run a `socket.chanrm('')` when passing in topic lists, let's definitely do that

lib/index.js
reqshark: > I don't have my heart set on 80 char line limits.

well I read code on the train from my phone..

reqshark:
I have mixed feelings about how we're handling the default string registration and then special `rmchan()` stuff when that's the case. it's taking me like 30 lines of code starting from around 100 and it seems overly complicated to me... but that's the simpliest way i saw solving it... 

actually maybe it's not that bad, i just took a second look after stepping away from the code...

@nickdesaulniers please review

test/tcp.js
```diff
@@ -15,6 +15,8 @@ test('tcp socket pub sub', function (t) {
     var pub = nano.socket('pub');
     var sub = nano.socket('sub');
 
+    sub.chan('');
+
```
nickdesaulniers: can be removed.

lib/index.js
```diff
@@ -372,6 +404,32 @@ function opt (option) {
   }
 };
 
+/* chan and rmchan sockopt methods. only relevant for subscription sockets */
+Socket.prototype.chan       = function(list){
```
nickdesaulniers: list here is an unused variable

lib/index.js
```diff
@@ -96,16 +96,48 @@ function Socket (type, opts) {
...
+
+  /*  subscription handling at initialization  */
+  if (opts.hasOwnProperty('chan')) {
+    if (Array.isArray(opts.chan)) {
+      opts.chan.forEach(register.bind(this));
+    } else {
+      throw new TypeError('chan requires an Array');
     }
+  } else {
+    if (type === 'sub') {
```
nickdesaulniers: this if can merge with the line above it `} else if (...) {`

lib/index.js
```diff
@@ -96,16 +96,48 @@ function Socket (type, opts) {
...
+      this.register(''); //default topic is an empty string
+      setImmediate(function(){ this.chanDefault = true; }.bind(this));
```
nickdesaulniers: What is this?

lib/index.js
```diff
@@ -96,16 +96,48 @@ function Socket (type, opts) {
...
+  function register(chan) {
+    if (this.chanDefault){
+      this.chanDefault = !this.chanDefault;
+      return this.rmchan('');
     }
+    if (nn.Chan(this.binding, nn.NN_SUB_SUBSCRIBE, chan) !== -1) {
+      return this.channels[chan] = true;
+    }
+    return this.emit('error', new Error( nn.Err() + ' : ' + chan));
+  }
```
nickdesaulniers: this functionality is definitely duplicated by `this.chan`

lib/index.js
```diff
@@ -372,6 +404,32 @@ function opt (option) {
...
 
+/* chan and rmchan sockopt methods. only relevant for subscription sockets */
+Socket.prototype.chan       = function(list){
+
+  var i = arguments.length;
+  while(i--) {
+    if (nn.Chan(this.binding, nn.NN_SUB_SUBSCRIBE, arguments[i]) !== -1) {
+      this.channels[arguments[i]] = true;
+    } else {
+      this.emit('error', new Error( er.op + er.chan + arguments[i] + '\n'));
```
nickdesaulniers: this error handler with throw an error, er.op and er.chan

lib/index.js
```diff
@@ -372,6 +404,32 @@ function opt (option) {
...
+}
+
+Socket.prototype.rmchan     = function(){
+  var i = arguments.length;
+  while(i--) {
+    if (this.channels[arguments[i]]) {
+      if (nn.Chan(this.binding, nn.NN_SUB_UNSUBSCRIBE, arguments[i]) !== -1) {
+        delete this.channels[arguments[i]];
+      } else {
+        this.emit('error', new Error( er.op + er.rmchan + arguments[i] + '\n'));
```
nickdesaulniers: this error handler will throw an error.  Removed the definitions but did not replace the usages.

lib/index.js
```diff
@@ -96,16 +96,48 @@ function Socket (type, opts) {
...
-
-    for(var sokopt in sol){
-        if(opts.hasOwnProperty(sokopt)) this[sokopt](opts[sokopt]);
+  this.binding = nn.Socket(this.af_domain, this.protocol);
+  this.queue = [];
+
+  /*  subscription filter control */
+  this.channels     = {};
+  this.register     = register;
+  this.chanDefault  = false;
```
nickdesaulniers: let's remove this, and use the ninja variable technique we discussed on irc.

test/chan.js
```diff
@@ -0,0 +1,69 @@
+var nano = require('..');
+var test = require('tape');
+
+test('adding and removing subscription channels', function (t) {
+
+  //t.plan(7);
```
nickdesaulniers: don't we need to list the number of plans?  if not, this can be removed

test/chan.js
```diff
@@ -0,0 +1,69 @@
...
+
+  function finish(msg){
+    t.equal(msg, msgs[msgqty]); //'foo world'
+    t.equal(msg, msgs[4]); //msgqty received is only 4
+    t.equal(msg, 'foo world'); //prove case 4 is m4, the fifth msg
+    sub.close();
+    pub.close();
+    t.end()
+  }
+});
```
nickdesaulniers: we should add a test to handle the default subscriptions, since we're doing extra logic there. more thoughts on that in a few mins.

test/chan.js
```diff
@@ -0,0 +1,69 @@
...
+  var msgs = ['hello world','hello world','hello world','bar world','foo world'];
+  var msgqty = -1; //starting here so first msg counted can be zero
+  var sent = 0;
+
+  sub.on('message', function (buf) {
+
+    var msg = String(buf);
+
+    //increment msgqty when a msg is received (happening about every 100ms)
+    //we'll call finish() on the 5th msg received (case 4, see line 68)
```
nickdesaulniers: I wouldn't recommend putting explicit line numbers, it's brittle and prone to breakage as the code changes.

test/chan.js
nickdesaulniers: The test should:
* show that constructing a socket without explicitly subscribing to any channel receives at least 2 messages of different prefixes.
* show that constructing a socket without explicitly subscribing to any channel during construction, but subscribes after construction using chan receives correctly prefixed messages but not others.
* show that constructing a socket explicitly subscribing to a channel during construction receives the appropriately prefixed messages but not others.
* show that constructing a socket explicitly subscribing to two or more channels during construction, later rmchan all but one, will stop receiving messages to channels no longer subscribed to.
* what happens when we rm all chans?

test/chan.js
reqshark: ok. yes, we either need to list plans or call `t.end()`

test/chan.js
reqshark: ok but

> * show that constructing a socket without explicitly subscribing to any channel receives at least 2 messages of different prefixes.

seems like redundant coverage in `test/ipc.js`, `test/inproc.js`, `test/tcp.js`

lib/index.js
reqshark: oof, i think i pushed the wrong state. haha it's on my home computer. anyway, `this.register` function iterates the list. i'll update that to reflect list as an array as discussed earlier

lib/index.js
reqshark: the `chanDefault` flag. 

waiting a turn of the event loop before setting `this.chanDefault` to true is a guarantee that `register()`'s check on line 122, occurs before `this.chanDefault` gets set to true.

it's excessive and we don't need `this.chanDefault`, if we set the empty string on `this.channels`, like we discussed on IRC this morning

lib/index.js
reqshark: ya i pushed the wrong version of the code, it's supposed to call `register()`

lib/index.js
reqshark: not the code i meant to push.

lib/index.js
reqshark: dito

src/node_nanomsg.cc
reqshark: `clang-format -style=Mozilla -i src/node_nanomsg.cc` approved!   :rocket:

reqshark:
sorry still WIP right now, I'll update with another note when it's ready for the final review..

lib/index.js
```diff
@@ -96,16 +96,42 @@ function Socket (type, opts) {
...
-    if(this.af_domain == nn.AF_SP) {
-        if (this.receiver) this._startPollReceive();
+  function register(chan) {
+    if (this.channels.hasOwnProperty('')) {
+      this.rmchan(''); return this.register(chan);
+    } else if (nn.Chan(this.binding, nn.NN_SUB_SUBSCRIBE, chan) !== -1) {
+      return this.channels[chan] = true;
     }
+    return this.emit('error', new Error( nn.Err() + ' : ' + chan));
+  }
```
nickdesaulniers: Can we make register a method of the prototype?
```js
Socket.prototype._register(chan) {};
```
Let's make each statement it's own line.
Let's remove return values since they're not used by any callers.
```js
if (this.chan...) {
} else if (nn.Chan) {
} else {}
// no explicit return statements
```

After this is fixed, the PR looks good to merge.  Now that Travis is hooked up, I trust it enough that I don't have to pull down the patch and test it locally! :rocket: :guitar:

lib/index.js
```diff
@@ -372,6 +398,27 @@ function opt (option) {
...
+  var i = arguments.length;
+  while(i--) {
+    if (this.channels[arguments[i]]) {
+      if (nn.Chan(this.binding, nn.NN_SUB_UNSUBSCRIBE, arguments[i]) > -1) {
+        return delete this.channels[arguments[i]];
+      }
+      return this.emit('error', new Error( nn.Err() + ' : ' + chan));
+    }
+  };
+}
```
nickdesaulniers: Actually, the return values here are also pretty useless, let's remove explicit return statements and just have the innermost if be an if-else.

nickdesaulniers:
> fixes #31 and alters our pubsub API, rendering it incompatible.

While it doesn't change the pubsub API, it does remove the getsockopt and setsockopt api, and thus will require a major version bump.

nickdesaulniers:
Let's focus on getting this wrapped up.  There's just one last line note with a few nits, then this is good to land! :rocket: :guitar:

reqshark:
ok let me just fix this merge conflict and add those last nits, I'll sort this out shortly

PR #69, opened 2015-03-18
update travis node versions

reqshark:
are we even running anything on the travis ci?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,10 @@
+branches:
+  only:
+    - master
 language: node_js
 node_js:
   - "0.10"
-  - "0.11"
+  - "0.12"
+  - "iojs"
+before_install: make
+script: make check
--- a/Makefile
+++ b/Makefile
@@ -5,10 +5,10 @@ ALL:
 	npm i
 
 check:
-	npm t
+	find test/*.js test/standalone/*.js | xargs -n 1 node | node_modules/tap-difflet/bin/tap-difflet
 
 test:
-	npm t
+	find test/*.js test/standalone/*.js | xargs -n 1 node | node_modules/tap-nyan/bin/cmd.js
 
 clean:
 	rm -fr build && rm -rf node_modules
--- a/package.json
+++ b/package.json
@@ -4,14 +4,15 @@
   "description": "Node bindings for nanomsg",
   "main": "lib/index.js",
   "dependencies": {
-    "bindings": "1.1.1",
-    "nan": "~1.6.2"
+    "bindings": "1.2.1",
+    "nan": "~1.7.0"
   },
   "devDependencies": {
     "esformatter": "0.0.16",
     "should": "~3.2.0-beta1",
     "tap-nyan": "0.0.2",
-    "tape": "~2.12.0"
+    "tape": "~2.12.0",
+    "tap-difflet":"0.3.0"
   },
   "scripts": {
     "test": "find test/*.js test/standalone/*.js | xargs -n 1 node | node_modules/tap-nyan/bin/cmd.js",
@@ -28,5 +29,7 @@
   "repository": {
     "type": "git",
     "url": "https://github.com/nickdesaulniers/node-nanomsg.git"
-  }
+  },
+  "author": "Nick Desaulniers <ndesaulniers@mozilla.com> (http://nickdesaulniers.github.io)",
+  "license": "MIT"
 }
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # nanomsg for node
 
-[![Build status](https://ci.appveyor.com/api/projects/status/07j7o9juuktas2uk)](https://ci.appveyor.com/project/tcr/node-nanomsg) [![Build Status](https://travis-ci.org/tcr/node-nanomsg.svg)](https://travis-ci.org/tcr/node-nanomsg)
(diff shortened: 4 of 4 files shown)


nickdesaulniers:
@tcr set this up, it looks like it's pulling from his fork: https://travis-ci.org/tcr/node-nanomsg

Can't say I know much about travis-ci.

reqshark:
cool, ya let's use this fork instead

nickdesaulniers:
please update the badge in the readme to:
```markdown
[![Build Status](https://travis-ci.org/nickdesaulniers/node-nanomsg.svg)](https://travis-ci.org/nickdesaulniers/node-nanomsg)
```

nickdesaulniers:
Should we also test older versions of node? or remove 0.10 from testing?  I think we could test all the way back to 0.6, but not have a hard requirement for supporting anything other than 0.12.  That way we can inform users in the readme if they need an explicit version of node.  Once we know, that would be good to put in the readme.

reqshark:
> not have a hard requirement for supporting anything other than 0.12

if we set out a hard requirement for constant support it should be latest `io.js` release. 

but I guess we could also target 0.12 and 0.10, I mean all these versions are the most important areas of coverage IMO, one is probably not necessarily more citical than any other. 

and if 0.6 actually passes then I'll buy you a beer.

nickdesaulniers:
> if we set out a hard requirement for constant support it should be latest io.js release. 

If someone had numbers showing me a significant drop in marketshare of node.js in favor of io.js, then yes I'd be fine with that.  Right now I'm not convinced it's anything more than a flash in the pan power play.  And it's not like we're bending over backwards here to support 0.12 as opposed to io.js.

https://news.ycombinator.com/item?id=9109499

reqshark:
we definitely get better native support in more recent versions, not even considering C++11 features.

for example [i found](https://github.com/reqshark/nanscope) running `nanscope()` impacts the performance of native context switches in `0.11`/`0.12` on methods where invoking the V8 scope handler is unnecessary, something that breaks support for `0.10` completely.

if experimenting with the compiler's taste for node versions is on the table, [let's just drop in some `#include` statements](https://github.com/reqshark/nanomsg.iojs/blob/master/lib/nanomsg.cc#L47-L51)

nickdesaulniers:
What do those headers buy us?  We're not using either?

reqshark:
> What do those headers buy us?

flexibility to exploit different flavors of native support across different node versions.

we get different native implementations of the same export method.

reqshark:
sorry those should have been named `.cc` instead of `.h` but since the compiler doesn't care neither do i

.travis.yml
```diff
@@ -1,4 +1,10 @@
+os:
+  - linux
+branches:
+  only:
+    - master
```
nickdesaulniers: will this no longer tests PR's on other branches?

package.json
```diff
@@ -1,32 +1,36 @@
...
-  "main": "lib/index.js",
+  "name":               "nanomsg",
+  "version":            "0.4.2",
+  "description":        "Node bindings for nanomsg",
+  "main":               "lib/index.js",
   "dependencies": {
-    "bindings": "1.1.1",
-    "nan": "~1.6.2"
+    "bindings":         "1.2.1",
+    "nan":              "~1.7.0"
```
nickdesaulniers: I'm ok with bumping these, but let's not rewrite the formatting of the entire file.

reqshark: what?! way easier to read this way, check it out in your text editor, you don't agree?

.travis.yml
reqshark: no PRs are always tested

if you commit to an open PR travis re-runs each commit

but let's say you were working on a branch and wanted to save some work... if you haven't PRed that yet then travis won't bother running and showing a bunch of extra test outcomes in the travis build area on their site, so we can suppress noise from builds we're not interested in looking at

package.json
nickdesaulniers: I'm indifferent; that's subjective and shifts git-blame.

reqshark: ya we should try to preserve empirical history that could explain unknowns down the road. 

so definitely, git-blame is prima facie > subjective discrepancies.  i'll revert that part

tcr:
:+1:  for switching the Travis CI branch to @nickdesaulniers'. I think it required I have collaborator access to this repo, so I used the fork back then.

Makefile
```diff
@@ -5,7 +5,7 @@ ALL:
 	npm i
 
 check:
-	npm t
+	-find test/*.js test/standalone/*.js | xargs -n 1 node | node_modules/tap-spec/bin/cmd.js
```
nickdesaulniers: woah, what's up with the leading hyphen? is that valid?

reqshark: to avoid linux edge case using `make`. hyphen suppresses an improper exit code, nice catch

nickdesaulniers: so it should have the hyphen?

reqshark: yep. this way travis displays a mocha style spec. 

there might be a better alternative for travis. 

we could run tap-spec as the default `npm test` and have something like `make nyan` for the preferred nyan-cat terminal output (unfortunately not helpful/intended for travis ci output).

reqshark:
landed e6a8e235e6e874d2059bd844a693fddac0bdd31e.

Issue #70, opened 2015-03-19
update socket.shutdown


reqshark:
@nickdesaulniers, we should probably implement this method [the way we say it's implemented](https://github.com/nickdesaulniers/node-nanomsg#socketshutdownaddress).

### socket.shutdown(*address*)

call `shutdown(str)` w/ an address string to detach the socket connected/bound to that address.

makes way more sense than [trying to cope with obscure return values](https://github.com/nickdesaulniers/node-nanomsg/blob/0a44b5ec46811a3941d8c563666270303c3c5ed9/lib/index.js#L280): libnanomsg's `how` integer returned by low level calls to [`nn_bind()`](http://nanomsg.org/v0.5/nn_bind.3.html) or [`nn_connect()`](http://nanomsg.org/v0.5/nn_connect.3.html).

I think we just need to tuck that weird implementation detail into a new object property on the JS socket and use the `delete` operator to clear it when `shutdown()` gets called. No biggie... 

I'll PR that one after lunch tomorrow and we won't have to worry about updating the docs :)

nickdesaulniers:
fixed in #71

PR #71, opened 2015-03-19
socket.shutdown api adjustments

reqshark:
follow up on #70

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -97,6 +97,8 @@ function Socket (type, opts) {
     }
 
   this.binding = nn.Socket(this.af_domain, this.protocol);
+  this.bound = {};
+  this.connected = {};
   this.queue = [];
 
   /*  subscription filter control */
@@ -224,12 +226,24 @@ Socket.prototype._register = function(chan){
  */
 
 Socket.prototype.bind = function (addr) {
-    return this._protect(nn.Bind(this.binding, addr));
+  return this.bound[addr] = this._protect(nn.Bind( this.binding, addr ));
 }
 
 Socket.prototype.connect = function (addr) {
-    return this._protect(nn.Connect(this.binding, addr));
+  return this.connected[addr] = this._protect(nn.Connect( this.binding, addr ));
 }
+
+Socket.prototype.shutdown = function (addr) {
+  var eid;
+  if (this.bound.hasOwnProperty(addr)) {
+    eid = this.bound[addr];
+    delete this.bound[addr];
+  } else if (this.connected.hasOwnProperty(addr)) {
+    eid = this.connected[addr];
+    delete this.connected[addr];
+  }
+  return this._protect(nn.Shutdown(this.binding, eid));
+};
 
 Socket.prototype.flush = function () {
     while(this.queue.length) {
@@ -275,10 +289,6 @@ Socket.prototype.getsockopt = function (level, option) {
 
 Socket.prototype.setsockopt = function (level, option, value) {
     return this._protect(nn.Setsockopt(this.binding, level, option, value));
-};
-
-Socket.prototype.shutdown = function (how) {
-    return this._protect(nn.Shutdown(this.binding, how));
 };
 
 /**
--- a/lib/index.js
+++ b/lib/index.js
@@ -238,11 +238,12 @@ Socket.prototype.shutdown = function (addr) {
   if (this.bound.hasOwnProperty(addr)) {
     eid = this.bound[addr];
     delete this.bound[addr];
+    return this._protect(nn.Shutdown(this.binding, eid));
   } else if (this.connected.hasOwnProperty(addr)) {
     eid = this.connected[addr];
     delete this.connected[addr];
+    return this._protect(nn.Shutdown(this.binding, eid));
   }
-  return this._protect(nn.Shutdown(this.binding, eid));
 };
 
 Socket.prototype.flush = function () {
--- a/test/bind.js
+++ b/test/bind.js
@@ -1,184 +1,148 @@
-// https://github.com/chuckremes/nn-core/blob/master/spec/nn_bind_spec.rb
-
-var nano = require('../');
-var nn = nano._bindings;
+var nano = require('..');
 var test = require('tape');
 
+test('map bind address eid for valid INPROC address', function (t) {
+  t.plan(1);
 
-test('bind returns non-zero endpoint number for valid INPROC address', function (t) {
-    t.plan(1);
+  var sock = nano.socket('pub');
+  sock.bind('inproc://some_address');
 
-    var sock = nano.socket('pub');
-    var rc = sock.bind('inproc://some_address');
+  if (sock.bound['inproc://some_address'] > -1) {
+    t.pass('valid INPROC bind');
+  } else {
+    t.pass('INPROC bind fail');
+  }
 
-    if(rc < 0) {
-        t.fail('INPROC endpoint number invalid');
-    } else {
-        t.pass('INPROC endpoint number valid');
-    }
+  sock.close();
+});
 
+test('map bind address eid for valid IPC address', function (t) {
+  t.plan(1);
+
+  var sock = nano.socket('pub');
+  sock.bind('ipc://some_address');
+
+  if (sock.bound['ipc://some_address'] > -1) {
+    t.pass('valid IPC bind');
+  } else {
+    t.pass('IPC bind fail');
+  }
+
+  sock.close();
+});
+
+test('map bind address eid for valid TCP address', function (t) {
+  t.plan(1);
+
+  var sock = nano.socket('pub');
+  sock.bind('tcp://127.0.0.1:5555');
+
+  if (sock.bound['tcp://127.0.0.1:5555'] > -1) {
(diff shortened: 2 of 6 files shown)


reqshark:
@tcr, @nickdesaulniers review? esp. changes to how bind and connect work

nickdesaulniers:
I don't see the value add here. :money_with_wings: :dollar: :dollar: :dollar:

reqshark:
this way shutdown is basically the opposite of `bind` or `connect`.

you call `bind` on an endpoint --> socket is bound.

you call `shutdown` on that endpoint --> socket is unbound.

this API fits more appropriately the philosophy of nanomsg, here's how connect is documented: 
> nn_connect - add a remote endpoint to the socket

now calling shutdown lets you remove an endpoint (rather than dealing with `eids`).

remember: the nanomsg socket can bind or connect to multiple heterogeneous endpoints as well as shutdown any of these established links.

lib/index.js
```diff
@@ -196,13 +197,31 @@ Socket.prototype._stopPollReceive = function () {
...
 }
 
 Socket.prototype.connect = function (addr) {
-    return this._protect(nn.Connect(this.binding, addr));
+  var eid = nn.Connect( this.binding, addr )
+  if(eid < 0){
+    return this.emit('error', new Error(nn.Err() +': '+this.type+' connect@' + addr));
+  }
+  this.how[addr] = [eid,'connect'];
+  return this.how; //socket's active endpoints
```
nickdesaulniers: instead of saving a tuple of the eid and a string, let's just use two different properties.  Example:

```js
this.connections = {};
this.binds = {};

Socket.prototype.bind = function (addr) {
  var eid = this._protect(nn.Bind(this.binding, addr));
  this.binds[addr] = eid;
  return eid;
};

ditto for connect.  then, we could also probably build from there fixes for #10 and #11, in another PR.
```

lib/index.js
```diff
@@ -196,13 +197,31 @@ Socket.prototype._stopPollReceive = function () {
...
 }
 
+Socket.prototype.shutdown = function (addr) {
+  var confirm = this.how[addr][1] + ' endpoint '+ addr + ' shutdown';
+  if(nn.Shutdown(this.binding, this.how[addr][0]) < 0){
+    return this.emit('error', new Error(nn.Err()+': '+this.type+' shutdown@'+addr));
+  }
+  delete this.how[addr]; return confirm;
+};
```
nickdesaulniers: ... then here it would be:
```js
if(Object.keys(this.binds).conatains(addr)) {
 ...
  delete this.binds[addr];
} else if (Object.keys(this.connects).contains(addr)) {
 ...
  delete this.connects[addr];
}
```

nickdesaulniers: Wow, that was stupid of me!
```js
if (addr in this.binds) {
} else if (addr in this.connects)
```
Otherwise we'd be scanning every key!

lib/index.js
reqshark: I'm good with that because it adds some transparency if you're just checking `console.log` on the socket quickly for bound and/or connected addresses.

but I'd rather call it:

```js
this.bound[addr] = eid;
```

lib/index.js
nickdesaulniers: or even:
```js
if (addr in this.binds || addr in this.connects) {
  ...
  delete this.binds[addr];
  delete this.connects[addr];
}
```
Though this doesn't let you detect which you shutdown from, though I'm not sure it's necessary.  Also, not sure if a socket would be both connected and bound to the same endopoint?

lib/index.js
nickdesaulniers: :rocket: :+1:

lib/index.js
reqshark: same thing, more natural like:
```js
delete this.connected[addr];
```

nickdesaulniers: :rocket: :rocket: :rocket:

reqshark: > not sure if a socket would be both connected and bound to the same endpoint?

these sockets are made for mixing up and doing new kinds of patterns

nickdesaulniers: > these sockets are made for mixing up and doing new kinds of patterns
I meant I think it makes no sense to do something like:

```js
var socket = nano.socket('pub');
var addr = 'tcp://127.0.0.1:12345';
// one socket, connected and bound to same ip addr
socket.connect(addr);
socket.bind(addr);
```

reqshark: correct: bind and connect on the same endpoint is a dead-end. thought you meant different endpoints I guess, idk why i wrote that

reqshark:
ok this is ready to review @nickdesaulniers?

lib/index.js
```diff
@@ -224,13 +226,37 @@ Socket.prototype._register = function(chan){
...
 }
 
 Socket.prototype.connect = function (addr) {
-    return this._protect(nn.Connect(this.binding, addr));
+  var eid = nn.Connect( this.binding, addr )
+  if(eid < 0){
+    return this.emit('error', new Error(nn.Err() +': '+this.type+' connect@' + addr));
+  }
+  this.connected[addr] = eid;
+  return this.connected; //socket's active endpoints
```
nickdesaulniers: I'm not found of duplicating functionality from `_protect` for the sake of custom error strings.  Let's ruse `protect`:
```js
Socket.prototype.connect = function (addr) {
  return (this.connected[addr] = this._protect(nn.Connect(this.binding, addr)));
};
```

lib/index.js
```diff
@@ -224,13 +226,37 @@ Socket.prototype._register = function(chan){
...
+    eid = this.bound[addr]; delete this.bound[addr];
+  } else if (this.connected.hasOwnProperty(addr)) {
+    eid = this.connected[addr]; delete this.connected[addr];
+  }
+
+  if (nn.Shutdown(this.binding, eid) < 0) {
+    this.emit('error', new Error(nn.Err()+': '+this.type+' shutdown@'+addr));
+  }
+  return eid;
+};
```
nickdesaulniers: Let's reuse `_protect`:
```js
if (...) {
  ...
  return this._protect(nn.Shutdown(this.binding, eid));
} else if (...) {
  ...
  return this._protect(nn.Shutdown(this.binding, eid));
}
```
And put each statement on its own line.

lib/index.js
nickdesaulniers: ditto for bind

lib/index.js
reqshark: > And put each statement on its own line.

kinda tough to argue that point

reqshark:
`_protect()` cleans up `bind()` and `connect()` code really well but we lose meaning on the err.

so compare that benefit and clean up against the loss of some err message precision:
> for the sake of custom error strings

see below, replacing these `expected:` err messages with the actual given by `_protect()`
 
```js
# bind exception: invalid TCP address (non-numeric port)
not ok 7 Invalid argument
  ---
    expected: 'Invalid argument: pub bind@tcp://127.0.0.1:port'
    actual:   'Invalid argument'
  ...
# bind exception: invalid TCP address (port out of range)
not ok 8 Invalid argument
  ---
    expected: 'Invalid argument: pub bind@tcp://127.0.0.1:65536'
    actual:   'Invalid argument'
  ...
# bind exception: unsupported transport
not ok 9 Protocol not supported
  ---
    expected: 'Protocol not supported: pub bind@zmq://127.0.0.1:6000'
    actual:   'Protocol not supported'
```
*also note: the lib's `nn_strerror()` numeric-to-string lookup degrades on osx, esp. some of these cases*

nickdesaulniers:
These would help if you were having tons of errors being emitted to narrow down one, but if you call socket.bind('zmq://...') for example and simply see 'protocol not supported', then it should be kind of obvious.

reqshark:
I was on the fence but I have to agree. :+1: 

will clean up JS operations elsewhere using more `_protect()`

lib/index.js
```diff
@@ -224,13 +226,25 @@ Socket.prototype._register = function(chan){
...
+Socket.prototype.shutdown = function (addr) {
+  var eid;
+  if (this.bound.hasOwnProperty(addr)) {
+    eid = this.bound[addr];
+    delete this.bound[addr];
+  } else if (this.connected.hasOwnProperty(addr)) {
+    eid = this.connected[addr];
+    delete this.connected[addr];
+  }
+  return this._protect(nn.Shutdown(this.binding, eid));
```
nickdesaulniers: I think this should go in both branches, otherwise, you're always invoking nn.Shutdown even when ei is undefined.

reqshark: so in other words if you pass the wrong addr to shutdown nothing happens.. sounds cool to me, i guess I was thinking an err when we get the address wrong wouldn't be bad.

nickdesaulniers:
oops forgot to squash and rebase and stuff, w/e

nickdesaulniers:
and forgot to add the `fixes #70`. man my game is off this morning

nickdesaulniers:
I just manually fixed that, but now GH's UI is confused as hell.

reqshark:
That's ok we'll fix it, actually looks fine to me! :)

Issue #74, opened 2015-03-26
remove unnecessary requires


nickdesaulniers:
all of the tests seem to include should, but dont seem to use it at all (unless importing should has side effects, looks like yes, not sure we use it still).

reqshark:
I try to take it out of my tests when I catch it. :+1:

nickdesaulniers:
I forgot to remove the dev dependencies

PR #75, opened 2015-03-26
check null terminated messages

nickdesaulniers:
https://github.com/nanomsg/nanomsg/pull/363#issuecomment-76783754

review? @reqshark

### DIFF
--- a/test/send.js
+++ b/test/send.js
@@ -25,3 +25,73 @@ test('send returns number of bytes queued for unbound socket', function (t) {
     sock.close();
 });
 
+test('send can take a string', function (t) {
+  t.plan(3);
+
+  var pub = nano.socket('pub');
+  var sub = nano.socket('sub');
+  var addr = 'inproc://some_address';
+  var msg = 'hi';
+
+  pub.bind(addr);
+  sub.connect(addr);
+
+  sub.on('message', function (buf) {
+    t.equal(buf.toString(), msg);
+    t.equal(buf.length, msg.length);
+    pub.close();
+    sub.close();
+  });
+
+  var bytes = pub.send(msg);
+  t.equal(bytes, msg.length);
+});
+
+test('send can take a buffer', function (t) {
+  t.plan(3);
+
+  var pub = nano.socket('pub');
+  var sub = nano.socket('sub');
+  var addr = 'inproc://some_address';
+  var msg = new Buffer('hello');
+
+  pub.bind(addr);
+  sub.connect(addr);
+
+  sub.on('message', function (buf) {
+    t.equal(buf.toString(), msg.toString());
+    t.equal(buf.length, msg.length);
+    pub.close();
+    sub.close();
+  });
+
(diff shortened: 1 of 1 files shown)


test/send.js
```diff
@@ -1,7 +1,5 @@
 // https://github.com/chuckremes/nn-core/blob/master/spec/nn_send_spec.rb
 
-var assert = require('assert');
-var should = require('should');
 var nano = require('../');
 var nn = nano._bindings;
```
reqshark: we can remove `var nn = nano._bindings;` here

nickdesaulniers: let's leave it for this PR and I'll target unnecessary requires in #74

nickdesaulniers:
huh, bunch of failed tests, I wonder if the utf8 is getting GC'd

reqshark:
osx is passing.

fails on my ubuntu server same as the tests.. hmm

src/node_nanomsg.cc
```diff
@@ -199,30 +199,24 @@ NAN_METHOD(Shutdown) {
...
-
-    const char *data = node::Buffer::Data(object);
-    input = new std::string(data, node::Buffer::Length(object));
-
+  const int flags = args[2].integer;
+  const char* data = NULL;
+  size_t length = 0;
+
+  // check if Local<v8::Value> is a v8::String
+  if (args[1]->IsString()) {
```
reqshark: what's a V8::String?

nickdesaulniers:
I'm quite certain that because I'm using args[1] directly, it's getting GC'd before libuv sends the message, so whatever is sent is completely random memory!  Will need to copy the string. :(

nickdesaulniers:
Also, this only seems to failing for strings, not buffers.

reqshark:
just check for buffer instance otherwise assume string and don't return undefined. that might fix

nickdesaulniers:
pushing another test, not holding my breath though.  I know we have no issues for buffers, but we definitely do for strings, since I believe we're adding the null byte (so this patch still doesn't address the whole point of the PR)

nickdesaulniers:
ah! NO_NULL_TERMINATION

nickdesaulniers:
yep, don't merge yet, I've got a bunch of ideas to clean this up

nickdesaulniers:
so, this whole thing might have been a waste of time.  I just added a test that sends a string, and sure enough it looks like we have nothing to worry about

nickdesaulniers:
that took some serious patch surgery to save just the tests.  These tests help show we don't send null terminated msgs from the application layer.  patch ready for review

reqshark:
LGTM.

btw, I think we may end up encountering some of this issue again in the near future when we introduce zero-copy send operations using the [nn_allocmsg stuff](http://nanomsg.org/v0.5/nn_allocmsg.3.html)

reqshark:
sizzurp 727f552fd0ec095c3beb02c235721e47a7e931f7

PR #76, opened 2015-03-26
remove unnecessary requires Fixes #74

### LINKED ISSUE #74: remove unnecessary requires

nickdesaulniers:
review? @reqshark


reqshark:
sizzzurp 381ab57d17a8fddf98cd7d12181d10e217601237

PR #80, opened 2015-04-11
Perf test option --string | --buffer

comick:
Hey guys. Curious to see perf difference between using strings of buffer.

Results using strings:
```
$ make perf      
node perf/local_lat.js tcp://127.0.0.1:5555 1 100000& node perf/remote_lat.js tcp://127.0.0.1:5555 1 100000 && wait
message size: 1 [B]
roundtrip count: 100000
average latency: 140.844 [us]
node perf/local_thr.js tcp://127.0.0.1:5556 1 100000& node perf/remote_thr.js tcp://127.0.0.1:5556 1 100000 && wait
message size: 1 [B]
message count: 100000
throughput: 95334 [msg/s]
throughput: 0.763 [Mb/s]
```
Results using buffers:
```
$ make perf      
node perf/local_lat.js tcp://127.0.0.1:5555 1 100000& node perf/remote_lat.js tcp://127.0.0.1:5555 1 100000 && wait
message size: 1 [B]
roundtrip count: 100000
average latency: 140.93 [us]
node perf/local_thr.js tcp://127.0.0.1:5556 1 100000& node perf/remote_thr.js tcp://127.0.0.1:5556 1 100000 && wait
message size: 1 [B]
message count: 100000
throughput: 129001 [msg/s]
throughput: 1.032 [Mb/s]
```

Buffers is faster as expected, but string is convenient.
Now I'm testing if the C++ approach to string to buffer conversion is faster that js string to buffer conversion (which might be because of JIT).

Meanwhile perf scripts improved. Default uses buffer.


comick:
To compare to reference nanomsg some warm-up phase is also needed using JS. Some code might not be JIT compiled within the end of the perf script, doing some test to better understand what's going on.

nickdesaulniers:
Those are interesting observations!  [Here is the relevant code](https://github.com/nickdesaulniers/node-nanomsg/blob/master/src/node_nanomsg.cc#L133-L145) that handles stings vs buffers.  I would have thought we're doing more work for buffers, but maybe not.  @reqshark can you review?

reqshark:
wow that's awesome, yep I'm looking at it tonight

perf/common.js
```diff
@@ -0,0 +1,18 @@
...
+    switch (msgType) {
+        case '--buffer':
+            buf = new Buffer(sz);
+            buf.fill('o');
+            break;
+        case '--string':
+            buf = new Array(sz + 1).join('o');
+            break;
+        default:
+            console.err('Buffer type not supported please use --buffer or --string.');
```
reqshark: can we say `Unspecified msg type not supported, please use --buffer or --string`

perf/common.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+exports.createMsg = function(msgType, sz) {
+    var buf;
```
reqshark: hey Michele, can we use double spaces for JS tabs, or basically half the tab size of this javascript? 

@nickdesaulniers, don't we want to implement that standard for JS?

perf/common.js
reqshark: also i guess this isn't required yet?

reqshark: oh wait I guess we're defaulting to buffer, now i understand.. that's really cool

perf/common.js
nickdesaulniers: eh, for tests/perf I don't care as much, and I haven't finished the JS formatting yet.

reqshark:
landed d1ab5a6a96abe3776c4cf80a309c0862402f7c29! thank you!

reqshark:
@comick since I was playing around with the git history here, you might have to re-fork after this, at least i would probably want to delete my fork and just re-fork it again if I were you.

Otherwise in the future if you want to rebase and squash commits before I merge, that would also fix merge conflicts or avoid re-forking... next time before I merge I'll try to give you the option to squash/rebase

nickdesaulniers:
This can be avoided by committing your patch to a new branch.  Another benefit is that should you decide to abandon your changes, you can either commit or stash then checkout master; no commitment.  I usually do:

```
git checkout -b descriptiveBranchName
write patch
git commit -am "blah Fixes issueNum"
git push -u origin descriptiveBranchName
```

then issue PR from descriptiveBranchName to nanomsg/master.

reqshark:
also if it's possible to rebase off an upstream branch inside your fork, you could just force push to master

reqshark:
yep, check this section on rebase in the io.js wiki: https://github.com/iojs/io.js/wiki/Contributing#rebase

note that this advice from io.js is preferable to github's: https://help.github.com/articles/syncing-a-fork/

reqshark:
also about re-forking, if you do that, just please save the work in this tree first: https://github.com/comick/node-nanomsg/tree/async

PR #81, opened 2015-04-12
simplify C++ send

nickdesaulniers:
let's see if this passes Travis, last time it did not.

### DIFF
--- a/src/node_nanomsg.cc
+++ b/src/node_nanomsg.cc
@@ -128,29 +128,14 @@ NAN_METHOD(Send) {
   NanScope();
 
   int flags = args[2].integer;
-  std::string *input;
 
-  if (node::Buffer::HasInstance(args[1]->ToObject())) {
-
-    v8::Handle<v8::Object> object = args[1]->ToObject();
-
-    const char *data = node::Buffer::Data(object);
-    input = new std::string(data, node::Buffer::Length(object));
-
+  if (node::Buffer::HasInstance(args[1])) {
+    ret(NanNew<Number>(nn_send(S, node::Buffer::Data(args[1]),
+                               node::Buffer::Length(args[1]), flags)));
   } else {
-
     utf8 str(args[1]->ToString());
-
-    input = new std::string(*str);
+    ret(NanNew<Number>(nn_send(S, *str, str.length(), flags)));
   }
-
-  v8::Local<v8::Number> bytes;
-
-  bytes = NanNew<Number>(nn_send(S, input->c_str(), input->length(), flags));
-
-  delete input;
-
-  ret(bytes);
 }
 
 NAN_METHOD(Recv) {
--- a/test/send.js
+++ b/test/send.js
@@ -95,3 +95,24 @@ test('should not null terminate when sending strings', function (t) {
 
 });
 
+test('send can take a number', function (t) {
+  t.plan(2);
+
+  var pub = nano.socket('pub');
+  var sub = nano.socket('sub');
+  var addr = 'inproc://some_address';
+  var msg = Math.pow(2, 42);
+
+  pub.bind(addr);
+  sub.connect(addr);
+
+  sub.on('message', function (buf) {
+    t.equal(buf.toString(), msg.toString());
+    pub.close();
+    sub.close();
+  });
+
+  var bytes = pub.send(msg);
+  t.equal(bytes, msg.length);
+});
+


nickdesaulniers:
interesting, this seems to hang in test/chan.js on travis and my linux box too.

nickdesaulniers:
this has to do exclusively with my handling of strings. will fix

nickdesaulniers:
review? @reqshark

nickdesaulniers:
hmm, one thing that might be bad, is this changes it from:

if is buffer
   ...
else
  convert to string

to:

isString
  ...
else
  convert to buffer

So passing an int might not work right... I should add a test.

nickdesaulniers:
yeah, let me change it so we check for a buffer, otherwise always cast to string.

nickdesaulniers:
review? @reqshark

reqshark:
ok ya I'll take a look tonight after reviewing @comick's perf stuff

reqshark:
removing lines of C/C++ in our send operation, this is great stuff! LGTM!

nickdesaulniers:
cool you merge it then

comick:
hi Guys, I was investigating performance using JS-side conversion of buffers and C++ side conversion. It happens that the different is not significant and the code can be even simpler:

```

NAN_METHOD(Send) {
  NanScope();

  int flags = args[2].integer;
  Local<Object> buf = args[1].As<Object>();

  char *data = node::Buffer::Data(buf);
  int data_len = node::Buffer::Length(buf);

  int bytes_sent = nn_send(S, data, data_len, flags);

  ret(NanNew<Number>(bytes_sent));
}
```
```js
 if (!Buffer.isBuffer(msg)) {
        msg = new Buffer(String(msg), 'utf8');
    }
```

nickdesaulniers:
heh, yeah we had changed the buffer/string detection from JS to C++ thinking it would be faster (without having the profiling stuff in place).  @comick , you're latest PR adds the ability to check this.  Can you post the results with the results before/after your patch (the one you posted in the previous comment) is applied?  I'm for having more logic in JS and less in C++, but the three of us can make a more informed decision with some profiling numbers.

comick:
Keeping the C++ code simple allows us to easily implement real improvements there (like async bind, connect). V8 JIT compiler is rather impressing and we can trust it. I compared implementation patched with snippet in my previous comment and results are comparable, follows median result of 10 make bench run with string and buffers:

String:
```
make bench
node perf/local_lat.js tcp://127.0.0.1:5555 10 1000& node perf/remote_lat.js tcp://127.0.0.1:5555 10 1000 && wait
message size: 10 [B]
roundtrip count: 1000
average latency: 204.693 [us]
node perf/local_thr.js tcp://127.0.0.1:5556 10 100000& node perf/remote_thr.js tcp://127.0.0.1:5556 10 100000 && wait
message size: 10 [B]
message count: 100000
throughput: 129976 [msg/s]
throughput: 10.398 [Mb/s]
```

Buffers:
```
make bench
node perf/local_lat.js tcp://127.0.0.1:5555 10 1000& node perf/remote_lat.js tcp://127.0.0.1:5555 10 1000 && wait
message size: 10 [B]
roundtrip count: 1000
average latency: 139.449 [us]
node perf/local_thr.js tcp://127.0.0.1:5556 10 100000& node perf/remote_thr.js tcp://127.0.0.1:5556 10 100000 && wait
message size: 10 [B]
message count: 100000
throughput: 127812 [msg/s]
throughput: 10.225 [Mb/s]
```

The only conclusion I can make is that there's not significant perf loss. For better concerns some bigger messages should be compared, but I would still go for simple C++ and smart JS for reasons at the beginning.

comick:
(in tests output --string was hardcoded in js files :) )

nickdesaulniers:
Can you edit that comment to show:

String [PR #81 patch]: ...
Buffer [PR #81 patch]: ...

String [https://github.com/nickdesaulniers/node-nanomsg/pull/81#issuecomment-92009674 patch]: ...
Buffer [https://github.com/nickdesaulniers/node-nanomsg/pull/81#issuecomment-92009674 patch]: ...

Because the throughput in [your second to last comment](https://github.com/nickdesaulniers/node-nanomsg/pull/81#issuecomment-92018981) is an order of magnitude higher [than from your PR](https://github.com/nickdesaulniers/node-nanomsg/pull/80), though it's 2am here and I can't spot the difference in the command line args, if there are any.

comick:
Here was "make bench", there was "make perf". They use different message sizes by default.
Follows results (median) using 1 byte messages with strings and buffers. 1 byte msg should make important possible overhead from conversion, but results are still comparable:
```
make strperf
node perf/local_lat.js tcp://127.0.0.1:5555 1 100000& node perf/remote_lat.js tcp://127.0.0.1:5555 1 100000 --string && wait
message size: 1 [B]
roundtrip count: 100000
average latency: 146.339 [us]
node perf/local_thr.js tcp://127.0.0.1:5556 1 100000& node perf/remote_thr.js tcp://127.0.0.1:5556 1 100000 --string && wait
message size: 1 [B]
message count: 100000
throughput: 127050 [msg/s]
throughput: 1.016 [Mb/s]
```
```
make bufperf
node perf/local_lat.js tcp://127.0.0.1:5555 1 100000& node perf/remote_lat.js tcp://127.0.0.1:5555 1 100000 --buffer && wait
message size: 1 [B]
roundtrip count: 100000
average latency: 140.307 [us]
node perf/local_thr.js tcp://127.0.0.1:5556 1 100000& node perf/remote_thr.js tcp://127.0.0.1:5556 1 100000 --buffer && wait
message size: 1 [B]
message count: 100000
throughput: 129001 [msg/s]
throughput: 1.032 [Mb/s]
```

reqshark:
>  larger messages result in larger throughput. 1B vs 10B.

expression of an important law in network physics IMO 

(it's not a real law i don't think haha, but to me it is.. :) and i wouldn't be surprised if there was academic stuff backing that up)

nickdesaulniers:
@comick , I think the string/buffer check being moved from C++ back to JS is worthwhile.  If you want, please issue a PR, with the results of profiling before __and__ after your patch is applied.

PR #83, opened 2015-04-12
simplify makefile

nickdesaulniers:
review? @reqshark


nickdesaulniers:
In gnu make, you can list target dependencies, and those will get run first in order.

reqshark:
:+1: :+1:

PR #85, opened 2015-04-12
No macros

nickdesaulniers:
review? @reqshark 

when you squash, can you reference issue number eighty four so we auto close it?


reqshark:
sure thing! 

cool, a lot of cleanup here

PR #86, opened 2015-04-12
polyfill buffer.equals Fixes #82

### LINKED ISSUE #82: bad test case
The test send.js says this on Travis in Node 0.10:

/home/travis/build/nickdesaulniers/node-nanomsg/node_modules/tape/index.js:75
        throw err
              ^
TypeError: Object hello has no method 'equals'
    at Socket.<anonymous> (/home/travis/build/nickdesaulniers/node-nanomsg/test/send.js:87:17)
    at Socket.emit (events.js:95:17)
    at Socket._receive (/home/travis/build/nickdesaulniers/node-nanomsg/lib/index.js:184:10)
    at Socket.<anonymous> (/home/travis/build/nickdesaulniers/node-nanomsg/lib/index.js:198:30)

I'm pretty sure that's from this: https://github.com/nickdesaulniers/node-nanomsg/blob/master/test/send.js#L87 when I removed should.

nickdesaulniers:
review? @reqshark

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,10 +8,11 @@
     "nan": "~1.7.0"
   },
   "devDependencies": {
+    "buffer-equals-polyfill": "1.0.0",
     "esformatter": "0.0.16",
+    "tap-difflet": "0.3.0",
     "tap-nyan": "0.0.2",
-    "tape": "~2.12.0",
-    "tap-difflet": "0.3.0"
+    "tape": "~2.12.0"
   },
   "scripts": {
     "test": "find test/*.js test/standalone/*.js | xargs -n 1 node | node_modules/tap-nyan/bin/cmd.js",
--- a/test/send.js
+++ b/test/send.js
@@ -2,9 +2,11 @@
 
 var nano = require('../');
 var nn = nano._bindings;
-
 var test = require('tape');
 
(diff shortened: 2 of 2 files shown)


nickdesaulniers:
do you think we should throw in case Buffer.prototype.equals is not defined, to prevent people from arbitrarily removing the polyfill?

reqshark:
> arbitrarily removing the polyfill

it seems extremely unlikely. how would that even happen when it's just a dep for the test?

I'm thinking: isn't `require('buffer-equals-polyfill')`, according to the node require algorithm getting called long after `Buffer` class/prototype?

reqshark:
is [this the github repo](https://github.com/davidmarkclements/buffer-equals-polyfill) for it?

nickdesaulniers:
yeah

reqshark:
eh, i wouldn't worry about that happening, because even if that could happen I think it falls back to some comparative logic in here: https://github.com/soldair/node-buffer-compare, which looks fine to me

reqshark:
hey this is a great fix for that test when using node 0.10!

reqshark:
sizzurp f08bdd09aee5ebafe1afbc98f231c70866fdcc04!

Issue #79, opened 2015-04-02
zero-copy


reqshark:
zero-copy becomes increasingly useful as msg size grows. 

for people not familiar with zero-copy or interested in more background info, [here's a nice article explaining what it is and why it's important](http://www.ibm.com/developerworks/library/j-zerocopy/).

libnanomsg supports zero-copy msg transport for both send and recv operations. Ok so when and where do we want to be doing *zero-copy* in nanomsg? I think according to the lib's creator Martin Sustrik, probably where: [large starts around 512kB, half a megabyte](http://linuxcnc.mah.priv.at/irc/%23nanomsg/2014-10-28.html#04:45:27). 

now since we need to enable zero-copy, I guess the only question is where and how in node-nanomsg. Right off the bat, I think a JS style sockopt with a `Boolean` would be a great way. 

at first i was thinking about an opt available to send, doing something like:
```js
socket.send(msg, *opt*)
```

but translating that to the EventEmitter's `onmessage` callback is awkward, since we're <strong>not doing</strong> explicit recv operations in `node-nanomsg`. so that's why I was thinking maybe:

```js
socket.zerocopy(true)
```

reqshark:
btw outlook is super busy this week for me. 

i think i could reasonably PR this feature next week, so if anyone wants to take a stab at it in the meantime, then by all means, would be happy to review or build on that PR!

reqshark:
this morning i was reading from *The Art of Unix Programming* and some lines stuck me from the section on <strong>Compactness and the Strong Single Center</strong>:

> [Compactness] is an often-overlooked strength of the Unix tradition. Many of its most effective tools are thin wrappers around a direct translation of some single powerful algorithm.

-Eric S. Raymond

IMO, that's exactly what we're trying to do here in the `src/node_nanomsg.cc` file.

reqshark:
@nickdesaulniers and i discussed this last night and we are going to take our time here basically. 

One thing we talked about on zero-copy is, first of all, it's kind of complicated to get right and for some common architectures this will have little to no impact. 

We are in no rush to implement this; however once we do, look out for a win in application performance probably mostly for off-cloud networked linux machines. Like dedicated servers or data centers with bare metal direct access to instruction sets/caches soldered in proximity to the system's core processor chips

@sustrik has mentioned before:
> that API is designed in such a way that it allows CPU-bypass in RDMA and other such super-efficient technologies

reqshark:
we tried this and it didn't really work.

PR #97, opened 2015-06-10
a zerocopy send method

reqshark:
finally got around to fixing #79 here. @nickdesaulniers review?

We're already receiving messages in the zero-copy style. Originally I thought we needed to figure out a way to do zero-copy `nn_recv` but it was accomplished.

This PR adds a higher performance alternative send operation for outbound messages in zero-copy fashion. The API is straightforward and only accepts node buffers.

```js
socket.zerocopySend(buf);
```

if you're sending a string, you'll need do something like:
```js
var buf = Buffer('my string');
socket.zerocopySend(buf);
```

By creating this as a buffer-object-only API we're avoiding extra checks we do in the ordinary send operation, resulting in a measurable and consistent latency reduction across the board. As message size increases beyond half a megabyte, so do the benefits of this alternative send method. 

I'll add that while researching, it became clear to me that the purpose of exposing a zero-copy send method in libnanomsg is to allow custom memory allocations, i.e. something faster than `memcpy`.

### DIFF
--- a/src/node_nanomsg.cc
+++ b/src/node_nanomsg.cc
@@ -131,9 +131,12 @@ NAN_METHOD(ZerocopySend) {
   int len = node::Buffer::Length(args[1]);
 
   void *msg = nn_allocmsg(len, 0);
-  memcpy(msg, node::Buffer::Data(args[1]), len);
-
-  NanReturnValue(NanNew<Number>(nn_send (s, &msg, NN_MSG, 0)));
+  if (!msg) {
+    return NanThrowError("zero-copy send failed: nn_allocmsg() returned NULL pointer");
+  } else {
+    memcpy(msg, node::Buffer::Data(args[1]), len);
+    NanReturnValue(NanNew<Number>(nn_send (s, &msg, NN_MSG, 0)));
+  }
 }
 
 NAN_METHOD(Recv) {
--- a/src/node_nanomsg.cc
+++ b/src/node_nanomsg.cc
@@ -117,11 +117,23 @@ NAN_METHOD(Send) {
 
   if (node::Buffer::HasInstance(args[1])) {
     NanReturnValue(NanNew<Number>(nn_send(
-        s, node::Buffer::Data(args[1]), node::Buffer::Length(args[1]), flags)));
+      s, node::Buffer::Data(args[1]), node::Buffer::Length(args[1]), flags)));
   } else {
     v8::String::Utf8Value str(args[1]->ToString());
     NanReturnValue(NanNew<Number>(nn_send(s, *str, str.length(), flags)));
   }
+}
+
+NAN_METHOD(ZerocopySend) {
+  NanScope();
+
+  int s = args[0]->Int32Value();
+  int len = node::Buffer::Length(args[1]);
+
+  void *msg = nn_allocmsg(len, 0);
+  memcpy(msg, node::Buffer::Data(args[1]), len);
+
+  NanReturnValue(NanNew<Number>(nn_send (s, &msg, NN_MSG, 0)));
 }
 
 NAN_METHOD(Recv) {
@@ -355,6 +367,7 @@ void InitAll(Handle<Object> exports) {
   EXPORT_METHOD(exports, Connect);
   EXPORT_METHOD(exports, Shutdown);
   EXPORT_METHOD(exports, Send);
+  EXPORT_METHOD(exports, ZerocopySend);
   EXPORT_METHOD(exports, Recv);
   EXPORT_METHOD(exports, Errno);
   EXPORT_METHOD(exports, PollSendSocket);
--- a/examples/zerocopy.js
+++ b/examples/zerocopy.js
@@ -1,0 +1,13 @@
+var nano = require('..');
+var pub = nano.socket('pub');
+var sub = nano.socket('sub');
+var buf = Buffer('hello from nanomsg!');
+
+pub.bind('tcp://127.0.0.1:55555');
+sub.connect('tcp://127.0.0.1:55555');
+sub.on('message', function(msg){ console.log(String(msg)) });
+
+setInterval(send, 100);
+function send(){
+  return pub.zerocopySend(buf);
+}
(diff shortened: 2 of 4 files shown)


src/node_nanomsg.cc
```diff
@@ -117,13 +117,25 @@ NAN_METHOD(Send) {
...
   }
 }
 
+NAN_METHOD(ZerocopySend) {
+  NanScope();
+
+  int s = args[0]->Int32Value();
+  int len = node::Buffer::Length(args[1]);
+
+  void *msg = nn_allocmsg(len, 0);
```
nickdesaulniers: nn_allocmsg could fail.  Need to check if the pointer returned is `NULL`.

src/node_nanomsg.cc
```diff
@@ -117,13 +117,25 @@ NAN_METHOD(Send) {
...
+  memcpy(msg, node::Buffer::Data(args[1]), len);
+
+  NanReturnValue(NanNew<Number>(nn_send (s, &msg, NN_MSG, 0)));
+}
```
nickdesaulniers: @gdamore, does a msg allocated with `nn_allocmsg` need to be freed with `nn_freemsg`?  It's not clear to me from the man pages.

src/node_nanomsg.cc
```diff
@@ -117,13 +117,25 @@ NAN_METHOD(Send) {
...
 }
 
+NAN_METHOD(ZerocopySend) {
+  NanScope();
+
+  int s = args[0]->Int32Value();
+  int len = node::Buffer::Length(args[1]);
+
+  void *msg = nn_allocmsg(len, 0);
+  memcpy(msg, node::Buffer::Data(args[1]), len);
```
nickdesaulniers: This is great, seeing this helps me think about how we might support zero copy send.

Having a function called ZerocopySend that contains a memcpy is killing me.  I wonder if we could change the interface to only work with a special object we returned that was backed by a call to nn_allocmsg.

I'm not a fan of only supporting buffers and not strings.  If people wanted to go out of their way to have faster sends, then we could make the interface not quite as nice and keep the nicer API (vanilla send) around, too.

For instance, I wonder if we could do something like:

```javascript
var nanomsg = require('nanomsg');
var msg = "hello world";
var buffer = nanomsg.nanomsgBackedBuffer(msg);
nanomsg.zeroCopySend(buffer);
```

hmm maybe not, since that still involves copying the string into the nn_allocmsg provided buffer.  I'll have to think about how we could avoid that memcpy.

My other thought is why don't we just always use this in vanilla send if it's faster?  If the answer is because we skip the conditional send; my response is we want to support strings or buffers for usability.

src/node_nanomsg.cc
gdamore: Yes.

gdamore: Oh, in this case, with NN_MSG, I believe nn_send() takes care of that implicitly.

gdamore:
This whole thing is exactly the opposite of a "zero copy" send.  Zero copy means something (to me at least) very explicit -- you don't copy the data.  Furthermore, in this particular case, this seems kind of silly, since all you've done is move the logic that nanomsg would do under the covers in nn_send() into your outer wrapper.  I don't think any savings is gained with this code.  (You can read src/core/global.c nn_send() to see that this is the case, although its a little more complex as it deals with scatter gather which you're avoiding here, but that's a tiny number of cycles -- like probably only one or two branch instructions saved per message.)

gdamore:
True zero copy would be much harder to obtain.  libnanomsg would have to give you a way to register a "done" function, so that it could call you to release your databuffer, and then you'd pass that along with the databuffer (perhaps as fields to the msg itself) to the send routine.   You'd then have to promise in your application code not to modify or free the buffer until libnanomsg called your completion routine.

In the grand scheme of things, this is complex, and for probably the most common applications -- those with small message sizes, probably of little value.  (Value increases if your malloc() implementation sucks, but then you should fix *that* problem.)

I'd be open to the idea of someone writing an in-library message cache to reduce pressure on the allocator.  Indeed, I've done this on mangos, where its even more important to avoid triggering gc related stalls.

To be honest, node programs that are single threaded (big event loop) probably will never have problems here since malloc() shouldn't have the usual lock contention problems if there is only a single application thread.

reqshark:
@gdamore these are great insights thank you! you've given us plenty to consider. One area today where messages tend to accumulate before initiating any allocations, potentially something eligible for a scatter/gather pattern happens during the flush function--a kind of prerequisite to our `nn_send()`, see e.g. https://github.com/nickdesaulniers/node-nanomsg/blob/master/lib/index.js#L249-L255

gdamore:
Admittedly I don't understand what flush does -- seems to be a nodejs thing.  The more I look at this, the more I think the libnanomsg model is actually the wrong abstraction for node.  It seems like it would be better to write a pure JS implementation of the SP protocols; understanding that this may be a somewhat non-trivial task.  (I wrote the bulk of mangos over the course of a weekend or so, but it's had probably a man week or two of additional effort invested since.)

nickdesaulniers:
> It seems like it would be better to write a pure JS implementation of the SP protocols

Originally it was, then significant contributions came in to write a native add on that worked with Node.js' default event loop.  Unfortunately, that leads to small inconsistencies where we have trouble detecting time outs.  I must admit, I do not have enough experience with libuv to understand the pollSocket stuff we have.

reqshark:
> a pure JS implementation of the SP protocols

heh, just about to post my thoughts on the same line. :+1: 

Based on what I know of nanomsg SP semantics, that would require quite a bit of wrapping/binding around kernel APIs directly.

The only "pure" Node.JS JavaScript networking API derives from the c runtime: [libuv](https://github.com/libuv/libuv), by virtue of the same native binding techniques here in this repo

gdamore:
Well, just one more reason I still consider JS mostly a toy, especially on the server side.  (Can you say 64-bit integers?)  That said, I'd think *someone* must want to be able to do native JS networking - I mean you'd like to be able to speak to databases, telephony, etc. etc.

nickdesaulniers:
> I still consider JS mostly a toy

gr8 b8 m8!

Anyways, going to close this.  Upon further discussion it looks true zero copy will be more involved.

Issue #103, opened 2015-08-10
ecosystem upgrades


reqshark:
## relevant ecosystem changes

- [ ] nan 2

#### beta 6 targets: 
- [x] windows
- [x] linx
- [x] osx

In addition to beta 6, let's use recently overhauled [native abstractions](https://github.com/nodejs/nan/issues/376) to compile against API breakage upstream in V8 4.4.63.26... like if we want to use iojs3. no more Handle, it's Local now. 

btw, @nickdesaulniers, iojs3 brings changes to node::buffer that I think we need to take a look at 

last week I was having some problems just getting nan2 off the ground but [that was NPM related](https://github.com/nodejs/nan/issues/405). I'll take another swing at it as soon as I can. In the meantime, PRs on any of these are most welcome :)

thelinuxlich:
+1

reqshark:
ok status update: got osx on beta6. building it on my machine running 10.11 El Capitan...

we're getting a few symbol test failures in `test/symbol.js` and `nn_device()` is killing the process. this is totally expected and @nickdesaulniers had to send a patch upstream earlier this year to libnanomsg in order to get our symbols test to pass, so upstream is again missing a couple symbols for export

here's the branch and i'll try to finish this up over the weekend for the other operating systems: https://github.com/nickdesaulniers/node-nanomsg/tree/beta6osx

*side note: the websocket stuff is compiling!* Take that snailzilla! :snail:

reqshark:
linux is done.

```bash
$ git clone https://github.com/nickdesaulniers/node-nanomsg nano && cd nano
$ git checkout beta6
$ make
$ make check
```

reqshark:
btw i'm going on vacation to Maui :palm_tree: starting tomorrow so if someone wants to do a PR for upgrading to nan 2, that would be awesome.  otherwise I'll get to it when i get back around mid sept

cc @kkoopa

Issue #102, opened 2015-08-10
nanomsg 0.6


ksaritek:
forked nanommsg is at 0.5 version(feb 26). When sould nanomsg be upgraded?

reqshark:
yea totally :loudspeaker: we're behind on that. it must get done. 

I thought I was going to have time to do it this weekend, eh... i guess I could do it next weekend

if you can't wait, send a PR

reqshark:
updated and published to npm a9350236cf98da3cd24ca69c18c96f528e73692e

PR #104, opened 2015-08-15
beta6 upgrade

reqshark:
@nickdesaulniers, the upgrade is done.

addresses #102 and half of #103

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -12,20 +12,19 @@
         ['OS=="mac"', {
             'includes': [
                 'macosx.gypi',
-            ]    
+            ]
         }],
         ['OS=="linux"', {
             'includes': [
                 'linux.gypi',
-            ]    
+            ]
         }],
         ['OS=="win"', {
             'includes': [
                 'win.gypi',
-            ]    
+            ]
         }],
       ],
-
       'include_dirs': [
         'deps/nanomsg/src',
         'deps/nanomsg/src/aio',
@@ -99,6 +98,7 @@
         'deps/nanomsg/src/protocols/utils/lb.c',
         'deps/nanomsg/src/protocols/utils/priolist.c',
         'deps/nanomsg/src/transports/utils/backoff.c',
+        'deps/nanomsg/src/transports/utils/base64.c',
         'deps/nanomsg/src/transports/utils/dns.c',
         'deps/nanomsg/src/transports/utils/iface.c',
         'deps/nanomsg/src/transports/utils/literal.c',
@@ -120,24 +120,36 @@
         'deps/nanomsg/src/transports/tcp/ctcp.c',
         'deps/nanomsg/src/transports/tcp/stcp.c',
         'deps/nanomsg/src/transports/tcp/tcp.c',
+        'deps/nanomsg/src/transports/tcpmux/atcpmux.c',
+        'deps/nanomsg/src/transports/tcpmux/btcpmux.c',
+        'deps/nanomsg/src/transports/tcpmux/ctcpmux.c',
+        'deps/nanomsg/src/transports/tcpmux/stcpmux.c',
+        'deps/nanomsg/src/transports/tcpmux/tcpmux.c',
+        'deps/nanomsg/src/transports/ws/aws.c',
+        'deps/nanomsg/src/transports/ws/bws.c',
+        'deps/nanomsg/src/transports/ws/cws.c',
+        'deps/nanomsg/src/transports/ws/sha1.c',
+        'deps/nanomsg/src/transports/ws/sws.c',
+        'deps/nanomsg/src/transports/ws/ws.c',
+        'deps/nanomsg/src/transports/ws/ws_handshake.c',
         'deps/nanomsg/src/devices/device.c',
       ],
     },
     {
       'target_name': 'node_nanomsg',
+      'dependencies': [ 'nanomsg', ],
       'cflags': [ '-Wall -Werror -Wno-unused' ],
       'cflags_cc': ['-fexceptions'],
       'conditions': [
         ['OS=="mac"', {
-          'ldflags': ['-ldtrace'],
-          'libraries': ['-ldtrace' ],
+          'ldflags': [ '-L<(PRODUCT_DIR)' ],
+          'libraries': [ '-L<(PRODUCT_DIR)' ],
         }],
         ['OS=="linux"', {
         }],
         ['OS=="win"', {
         }],
       ],
-      'dependencies': [ 'nanomsg', ],
       'include_dirs': [
         "<!(node -e \"require('nan')\")",
       ],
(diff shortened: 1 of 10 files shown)


thelinuxlich:
great!

binding.gyp
```diff
@@ -120,24 +120,36 @@
...
       'target_name': 'node_nanomsg',
+      'dependencies': [ 'nanomsg', ],
       'cflags': [ '-Wall -Werror -Wno-unused' ],
       'cflags_cc': ['-fexceptions'],
       'conditions': [
         ['OS=="mac"', {
-          'ldflags': ['-ldtrace'],
-          'libraries': ['-ldtrace' ],
+          'ldflags': [ '-L<(PRODUCT_DIR) -ldtrace' ],
+          'libraries': [ '-ldtrace', '-L<(PRODUCT_DIR)' ],
```
nickdesaulniers: we don't mean to be linking in dtrace, right?

macosx.gypi
```diff
@@ -3,23 +3,30 @@
...
+            'NN_USE_KQUEUE',
             'NN_USE_IFADDRS',
+            'NN_USE_EVENTFD',
+            'NN_HAVE_MSG_CONTROL'
         ],
         'include_dirs': [
           'deps/nanomsg/src',
         ],
+        'cflags': [
+          '-Wall -Werror -Wno-unused -g -O2 -D_THREAD_SAFE'
```
nickdesaulniers: Don't need debug symbols.  Can we crank optimizations up to 11...I mean O3? :rocket:

binding.gyp
reqshark: nope, i'll take it out

macosx.gypi
reqshark: good idea!

macosx.gypi
```diff
@@ -3,23 +3,30 @@
...
+            'NN_USE_KQUEUE',
             'NN_USE_IFADDRS',
+            'NN_USE_EVENTFD',
+            'NN_HAVE_MSG_CONTROL'
         ],
         'include_dirs': [
           'deps/nanomsg/src',
         ],
+        'cflags': [
+          '-g -O3 -D_THREAD_SAFE'
```
nickdesaulniers: drop `-g`, we don't need debug symbols :oncoming_police_car:

nickdesaulniers:
I get an error with one test:
```
git submodule sync
git submodule update
npm i
npm t
...
/Users/Nicholas/code/node/nanomsg/node_modules/tape/index.js:75
        throw err
                  ^
Error: stream.push() after EOF
    at readableAddChunk (_stream_readable.js:149:15)
    at Socket.Readable.push (_stream_readable.js:126:10)
    at Socket.<anonymous> (/Users/Nicholas/code/node/nanomsg/lib/index.js:207:24)
```

reqshark:
are you sure you're pulling from the right submodule? what's the last commit hash on that submodule?

nickdesaulniers:
`* 5515f96 - (HEAD, tag: 0.6-beta) Bump ABI version for 0.6-beta. (4 weeks ago) <Garrett D'Amore>`

reqshark:
you're using node 0.12 aren't you

reqshark:
using node 0.12 i'm able to reproduce your result, apparently returning after the survey test passes (and just before it completes).

```bash
✔  bent@osx /Users/bent.cardan/reqshark/node-nanomsg  [beta6|✔ ] 
12:53 $ node test/survey
TAP version 13
# inproc socket survey
ok 1 (unnamed assert)
ok 2 (unnamed assert)
ok 3 (unnamed assert)
ok 4 (unnamed assert)
/Users/bent.cardan/reqshark/node-nanomsg/node_modules/tape/index.js:75
        throw err
              ^
Error: stream.push() after EOF
    at readableAddChunk (_stream_readable.js:149:15)
    at Socket.Readable.push (_stream_readable.js:126:10)
    at Socket.<anonymous> (/Users/bent.cardan/reqshark/node-nanomsg/lib/index.js:207:24)
```

we need to rewrite node.js survey sockets in light of updates to respondent/survey behavior in libnanomsg. note that what we're testing here calls
```js
socket.survey(msg)
``` 
on a surveyor socket, which is different than calling
```js
socket.send(msg)
```

this is [an inconsistent area of our API](https://github.com/nickdesaulniers/node-nanomsg/blob/b60a5130f7acbc03bacba57faf33212f88d0194a/lib/index.js#L311-L328) that has little or nothing to do with the upgrade. 

one problem i see here is that we're [removing a listener](https://github.com/nickdesaulniers/node-nanomsg/blob/b60a5130f7acbc03bacba57faf33212f88d0194a/lib/index.js#L323) that doesn't exist. I think this type specific API stuff was done back when libnanomsg only partially implemented the surveyor/respondent pattern.

since then the pattern's been improved by the clib, certainly in beta6 and possibly earlier in beta5. let's open a separate issue to track this and not let it prevent the upgrade

reqshark:
hmm, wait why does the survey pass on node 0.12 before the upgrade... 
```js
15:35 $ node test/survey.js 
TAP version 13
# inproc socket survey
ok 1 (unnamed assert)
ok 2 (unnamed assert)
ok 3 (unnamed assert)
ok 4 (unnamed assert)

1..4
# tests 4
# pass  4

# ok
```

that means it has something to do with the upgrade and we'll need to revise our survey socket code

lib/index.js
```diff
@@ -184,7 +184,7 @@ Socket.prototype._receive = function () {
             this._stopPollSend();
             this._stopPollReceive();
             this.emit('survey-timeout');
-            return;
+            return null;
```
reqshark: returning `null` here inside the survey's `send` operation ends of the stream properly instead of getting `EOF`

reqshark:
@nickdesaulniers i went through our survey code and made a small change to how survey's recv returns.

now I think the problem had more to do with streams than survey actually, though it's still interesting to me that we only saw this behavior with the upgrade.

anyway, by pushing a value of `undefined` onto a stream we get unexpected behavior. 

by pushing `null` we end the stream, which is what i think we intend to do here: https://github.com/nickdesaulniers/node-nanomsg/blob/beta6/lib/index.js#L187

reqshark:
dude when you get back from the BrazilJS conf Porto Alegre, this PR needs some :heart: ... after this is merged we need to fix iojs3.. nan 2... i was holding off on nan2 since this changes the clib slightly.

btw hope ur having a good time down there.. :palm_tree: this comment doesnt have anything to do with the PR but i was checking out your [your conference slides](http://nickdesaulniers.github.io/BrazilJS2015/slides/index.html) and was wondering if you could pls explain:

> Prefer task level parallelism (postMessage) to data level parallelism (SharedArrayBuffer)

nickdesaulniers:
k let's do :beers: brother!

nickdesaulniers:
@reqshark , can you publish to npm? :fried_shrimp:

reqshark:
done. now npm install has beta6 for everything < iojs-v2.5.0. 

We'll need to get nan2 updates done for v3+.

PR #108, opened 2015-09-04
Update to NAN 2.0.8

### LINKED ISSUE #103: ecosystem upgrades

kkoopa:
Fixes #103

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "lib/index.js",
   "dependencies": {
     "bindings": "1.2.1",
-    "nan": "~1.8.x"
+    "nan": "~2.0.8"
   },
   "devDependencies": {
     "buffer-equals-polyfill": "1.0.0",
--- a/src/node_nanomsg.cc
+++ b/src/node_nanomsg.cc
@@ -16,7 +16,6 @@
 using v8::Array;
 using v8::Function;
 using v8::FunctionTemplate;
-using v8::Handle;
 using v8::Local;
 using v8::Number;
 using v8::Object;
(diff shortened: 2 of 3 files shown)


nickdesaulniers:
wow! The PR in which @kkoopa basically rewrites the entire C++ part of this lib!  Thanks for the awesome PR.  The `FromJust` method name makes me cringe, but this looks good!  Let me verify the tests and merge!

kkoopa:
It's easy when one wrote the binding layer in the first place. Besides, the logic remains unchained, only mechanical search and replace was needed. Took me around ten minutes.

`FromJust` is [standard V8 nomenclature](https://v8docs.nodesource.com/io.js-3.0/d9/d4b/classv8_1_1_maybe.html#ad97c01315c73bf49b522c2996638a2d2) and is named so since it is paired with [`FromMaybe`](https://v8docs.nodesource.com/io.js-3.0/d9/d4b/classv8_1_1_maybe.html#ab0d2cdc887e031a2f60eec7e248c05d7), which should be used when there is a chance of the value being empty.

reqshark:
@kkoopa mahalo!

nickdesaulniers:
[I understand the concept of Maybe](http://nickdesaulniers.github.io/blog/2013/05/07/rust-pattern-matching-and-the-option-type/), but what is `Just`?

kkoopa:
It is just a value, that is, the opposite of nothing.

reqshark:
BTW looks our latest module release `2.1.1` also works with `node v4.0.0`.

excellent work @kkoopa, thanks again.

PR #113, opened 2015-09-10
run formatter on C++

nickdesaulniers:
we should have done this before merging the last change to these files

review? @reqshark


reqshark:
ok.. let's slaughter @kkoopa's git blame but other than that LGTM.

PR #114, opened 2015-09-10
rewrite bind/connect and remove setTimeouts from tests

### LINKED ISSUE #110: streams test timeout
on the travis iojs test yesterday and once today already on the node v4.0.0 test I was getting a timeout on our: `pipe a thousand msgs between incompatible socket types`

https://github.com/nickdesaulniers/node-nanomsg/blob/master/test/streams.js#L4 

i had to manually restart that otherwise it was just hanging so.. 

let's keep an eye on that travis CI timeout happening again and consider lowering the number of msgs to 300 or 500 or something, that would be a little easier for the travis box to handle during high load from everyone on github running their tests over there

reqshark:
fixes #110

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -322,7 +322,7 @@ Socket.prototype.flush = function () {
     this._stopPollSend();
 };
 
-Socket.prototype.close = function (fn) {
+Socket.prototype.close = function () {
     if(!this.closed) {
         // Prevent "Bad file descriptor" from recursively firing "error" event
         this.closed_status = nn.Close(this.binding);
--- a/lib/index.js
+++ b/lib/index.js
@@ -96,10 +96,14 @@ function Socket (type, opts) {
             break;
     }
 
-  this.binding = nn.Socket(this.af_domain, this.protocol);
-  this.bound = {};
-  this.connected = {};
-  this.queue = [];
+  this.binding      = nn.Socket(this.af_domain, this.protocol);
+
+  /*  track bound and connected sockets */
+  this.bound        = {};
+  this.connected    = {};
+
+  /*  track and flush msgs to libnanomsg */
+  this.queue        = [];
 
   /*  subscription filter control */
   this.channels     = {};
@@ -241,12 +245,60 @@ Socket.prototype._read = function (n) {}
  * Socket API
  */
 
-Socket.prototype.bind = function (addr) {
-  return this.bound[addr] = this._protect(nn.Bind( this.binding, addr ));
+Socket.prototype.bind = function (addr, fn) {
+  if (fn) return asyncBindWrapper.call(this, nn.asyncBind);
+  var ret;
+  this.sync('Bind', this, addr, ret);
+  return this.bound[addr];
+  function asyncBindWrapper (asyncBind) {
+    asyncBind(this.binding, addr, cb.bind(this));
+  }
+  function cb (eid) {
+    // set `this.bound[addr]` eid before cb fires
+    this.bound[addr] = eid;
+    return fn(eid);
+  }
 }
 
-Socket.prototype.connect = function (addr) {
-  return this.connected[addr] = this._protect(nn.Connect( this.binding, addr ));
+Socket.prototype.connect = function (addr, fn) {
+  if (fn) return asyncConnectWrapper.call(this, nn.asyncConnect);
+  var ret;
+  this.sync('Connect', this, addr, ret);
+  return this.connected[addr];
+  function asyncConnectWrapper (asyncConnect) {
+    asyncConnect(this.binding, addr, cb.bind(this));
+  }
+  function cb (eid) {
+    // set `this.connected[addr]` eid before cb fires
+    this.connected[addr] = eid;
+    return fn(eid);
+  }
+}
+
+/*  deasync helpers */
+Socket.prototype.sync = deasync(block);
+function block (nnMethod, ctx, opt, ret, done) {
+  var throttle = setInterval(function(){
+    if (ret !== undefined) {
+      clearInterval(throttle);
+      done();
+    }
+  }, 10);
+  ret = ctx._protect(nn[nnMethod]( ctx.binding, opt ));
+  switch (nnMethod) {
+    case 'Bind'     : return ctx.bound[opt]       = ret;
+    case 'Connect'  : return ctx.connected[opt]   = ret;
+  }
+}
+function deasync (fn) {
+  return function () {
+    var done=false, err, res;
+    fn.apply(this, Array.prototype.slice.apply(arguments).concat(cb));
+    while (!done) nn.block();
+    if (err) throw err;
+    return res;
+    function cb (e, r) { err=e; res=r; done=true; };
+  }
 }
 
 Socket.prototype.shutdown = function (addr) {
@@ -270,7 +322,7 @@ Socket.prototype.flush = function () {
     this._stopPollSend();
 };
 
-Socket.prototype.close = function () {
+Socket.prototype.close = function (fn) {
     if(!this.closed) {
         // Prevent "Bad file descriptor" from recursively firing "error" event
         this.closed_status = nn.Close(this.binding);
--- a/lib/index.js
+++ b/lib/index.js
@@ -291,7 +291,7 @@ function block (nnMethod, ctx, opt, ret, done) {
 function deasync (fn) {
   return function () {
     var done=false, err, res;
-    fn.apply(this, Array.prototype.slice.apply(arguments).concat(cb));
+    fn.apply(this, [].slice.call(arguments).concat(cb));
     while (!done) nn.block();
     if (err) throw err;
     return res;
--- a/lib/index.js
+++ b/lib/index.js
@@ -247,8 +247,7 @@ Socket.prototype._read = function (n) {}
 
 Socket.prototype.bind = function (addr, fn) {
   if (fn) return asyncBindWrapper.call(this, nn.asyncBind);
-  var ret;
-  this.sync('Bind', this, addr, ret);
+  this.sync('Bind', this, addr, undefined);
   return this.bound[addr];
   function asyncBindWrapper (asyncBind) {
     asyncBind(this.binding, addr, cb.bind(this));
@@ -262,8 +261,7 @@ Socket.prototype.bind = function (addr, fn) {
 
 Socket.prototype.connect = function (addr, fn) {
   if (fn) return asyncConnectWrapper.call(this, nn.asyncConnect);
-  var ret;
-  this.sync('Connect', this, addr, ret);
+  this.sync('Connect', this, addr, undefined);
   return this.connected[addr];
   function asyncConnectWrapper (asyncConnect) {
     asyncConnect(this.binding, addr, cb.bind(this));
--- a/src/node_nanomsg.cc
+++ b/src/node_nanomsg.cc
@@ -87,6 +87,70 @@ NAN_METHOD(Connect) {
   info.GetReturnValue().Set(Nan::New<Number>(nn_connect(s, *addr)));
 }
 
+class BindThread : public Nan::AsyncWorker {
+public:
+  BindThread(Nan::Callback *callback, int s, char *addr)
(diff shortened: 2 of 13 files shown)


nickdesaulniers:
still had a failure: https://travis-ci.org/nickdesaulniers/node-nanomsg/jobs/79752651

I don't think it's that we're running too many iterations of the tests, but that something is actually getting hung up.  Maybe add some printf debugging and push to travis to see what's going wrong there?

reqshark:
yea I wonder could a stream remain open after a `socket.close()` or visa versa? 

might need to look closer into how we end streams in the context of a closed socket

reqshark:
it seems like a test issue rather than a specific `socket.close()` issue. I double checked our implementation, and confirm we're emitting a `close()` event when the nanomsg socket closes. 

I think it's a separate issue, but we could slightly improve the implementation by holding off on emitting `close()`, needed to satisfy the node `Stream`, until lib returns properly from socket's `nn_close()`

What I think is happening here is a race condition when you close the same stream/socket firing the function/call also responsible for ending the test at a later point in the control flow of the function.  

test needs to end <strong>*before*</strong> we close the socket

test/streams.js
```diff
@@ -3,34 +3,35 @@ var test = require('tape');
...
 
-  pub.bind('tcp://127.0.0.1:64999');
-  sub.connect('tcp://127.0.0.1:64999');
+  pub.bind('tcp://127.0.0.1:34999');
+  sub.connect('tcp://127.0.0.1:34999');
 
-  pull.bind('tcp://127.0.0.1:65000');
-  push.connect('tcp://127.0.0.1:65000');
+  pull.bind('tcp://127.0.0.1:35000');
+  push.connect('tcp://127.0.0.1:35000');
```
nickdesaulniers: if you keep the old ports, and preincrement ++recv, does it fix the test?  I would expect one or the other, not both.  But we could play it safe and always use unique sockets per test.

reqshark: yea I think so, we need more unique ports. This could all be a red herring: probably just the test running right before stream causing idiosyncratic timeouts. 

but it's cool we're improving the `socket.close()` implementation :+1:

reqshark:
apparently, at times, it can take upwards of a hundred milliseconds or more for a socket to connect over local interfaces on travis boxes. 

that data center must be saturated with network activity or perhaps distributing access to local interfaces on virtual machines across some geographic distance, whatever the reason, allowing for the extra latency seems to resolve these idiosyncratic timeouts/failures we were getting.

we really should have async callbacks that fire after bind/connect to avoid all the wild guessing we're doing with `setTimeout`... but until then this will have do. 

@nickdesaulniers please review/merge

nickdesaulniers:
is this ready for review?

src/node_nanomsg.cc
```diff
@@ -37,7 +39,11 @@ NAN_METHOD(Close) {
     rc = nn_close(s);
   } while (rc < 0 && errno == EINTR);
 
-  info.GetReturnValue().Set(Nan::New<Number>(rc));
+  v8::Local<v8::Number> ret = Nan::New<Number>(rc);
+  Local<Value> argv[] = { ret };
```
nickdesaulniers: why are we creating an array of Local<Value>s?

src/node_nanomsg.cc
```diff
@@ -37,7 +39,11 @@ NAN_METHOD(Close) {
...
+
+  info.GetReturnValue().Set(ret);
+  callback->Call(1, argv);
```
nickdesaulniers: This doesn't make anything async, you're still calling the callback synchronously.  You'll need to coordinate with libuv to get notified when bind finishes.  See the work @comick did: https://github.com/comick/node-nanomsg/commit/87b2c5300b1a449b891c05fa2687690aec88731f .  Not sure if that's the correct way to interface with the event loop, but nn_bind is a blocking synchronous function, so you either have to spawn a thread and call it or work with libuv.

reqshark: That's a good explanation for why this was not fixing the timeout issue on Travis

reqshark:
Thanks for the code review. As you pointed out on the diff, this is not ready for merge. I need to review @comick's work and think about how best to do the async stuff.

src/node_nanomsg.cc
reqshark: That's the standard way to fill values given to v8 callbacks, I think we have to create this array even if we're going to only return one value as the parameter to the callback

reqshark:
@nickdesaulniers ok here's potentially two separate solutions to the timeout problem, and i think this might be what we actually want here:

* when bind/connect have only a single parameter passed, `addr`, do a complete synchronous block of the UV event loop. A new `uv_run()` with `UV_RUN_ONCE` option forces the *<strong>surrounding</strong>* uv context where it is called to block. In effect this creates a runloop pause that can be wrapped into its own callback, lending us the power to resume the uv runloop when a given completion flag is set to `true`, similar to how ES6 yield works. I first began to experiment with some of this dasync stuff a while back; see e.g, [how we can use it to create a blocking `sleep()` function in JS](https://github.com/reqshark/stall) *

* start to draft an async bind and connect, triggered when a 2nd callback param is passed by opening a v8 thread that returns the callback after the bind/connect work has been completed.

**inspired/plagarized from [the original deasync project](https://github.com/vkurchatkin/deasync)*. 

can i get another review?

nickdesaulniers:
cool, this will take me some time to review. I'll get to it this week.  Thanks for the hard work!

lib/index.js
```diff
@@ -270,7 +322,7 @@ Socket.prototype.flush = function () {
     this._stopPollSend();
 };
 
-Socket.prototype.close = function () {
+Socket.prototype.close = function (fn) {
```
nickdesaulniers: this argument is not being used?

reqshark: trizz

src/node_nanomsg.cc
```diff
@@ -87,6 +87,70 @@ NAN_METHOD(Connect) {
...
+      : Nan::AsyncWorker(callback), s(s), addr(addr) {}
+  ~ConnectThread() {}
+
+  // Executed inside worker-thread.
+  void Execute() {
+    int connected = nn_connect(s, addr);
+  }
+  void HandleOKCallback() {
+    Nan::HandleScope scope;
+    Local<Value> argv[] = { Nan::New<Number>(connected) };
```
nickdesaulniers: How is `connected` defined in Execute, but accessed in `HandleOKCallback`? wat :fire:

src/node_nanomsg.cc
```diff
@@ -87,6 +87,70 @@ NAN_METHOD(Connect) {
...
+  int connected;
+};
+
+NAN_METHOD(asyncConnect) {
+  Nan::Callback *callback = new Nan::Callback(info[2].As<Function>());
+
+  int s = Nan::To<int>(info[0]).FromJust();
+  String::Utf8Value addr(info[1]);
+
+  Nan::AsyncQueueWorker(new ConnectThread(callback, s, *addr));
```
nickdesaulniers: who calls `delete` on the heap allocated `ConnectThread`?

lib/index.js
```diff
@@ -241,12 +245,60 @@ Socket.prototype._read = function (n) {}
...
+  switch (nnMethod) {
+    case 'Bind'     : return ctx.bound[opt]       = ret;
+    case 'Connect'  : return ctx.connected[opt]   = ret;
+  }
+}
+function deasync (fn) {
+  return function () {
+    var done=false, err, res;
+    fn.apply(this, Array.prototype.slice.apply(arguments).concat(cb));
+    while (!done) nn.block();
```
nickdesaulniers: If `block` is implemented in C++, we should need to spin in a loop here.

nickdesaulniers: is implemented *correctly*

src/node_nanomsg.cc
reqshark: haha good one :black_joker:

lib/index.js
nickdesaulniers: *should not need*

ugh goes to bed :credit_card: :goberserk:

lib/index.js
```diff
@@ -241,12 +245,60 @@ Socket.prototype._read = function (n) {}
  * Socket API
  */
 
-Socket.prototype.bind = function (addr) {
-  return this.bound[addr] = this._protect(nn.Bind( this.binding, addr ));
+Socket.prototype.bind = function (addr, fn) {
+  if (fn) return asyncBindWrapper.call(this, nn.asyncBind);
+  var ret;
+  this.sync('Bind', this, addr, ret);
```
nickdesaulniers: I don't think we need `ret` in `bind` or `connect`.  We're not using it after invoking `this.sync`.  As far as I can tell, it's `undefined`.  I'm quite sure you can only use Arrays or Objects as in out parameters (reference types).

lib/index.js
reqshark: TBH I'm not sure how much spin is happening there. 

Just some common sorcery :dizzy: but definitely not deep magic :milky_way:

lib/index.js
reqshark: yea that's true `ret` must be undefined. So we could withhold the arg from being passed to `this.sync` and that should make it *just* `undefined` (at least declared) inside `this.sync`. 

it would work but look again at deasync helpers. `this.sync` is a transformed `function block()`, after getting passed to `deasyc(block)`.
```js 
/*  deasync helpers */
Socket.prototype.sync = deasync(block);
function block (nnMethod, ctx, opt, ret, done) {
  var throttle = setInterval(function(){
    if (ret !== undefined) {
      clearInterval(throttle);
      done();
    }
  }, 10);
  ret = ctx._protect(nn[nnMethod]( ctx.binding, opt ));
  switch (nnMethod) {
    case 'Bind'     : return ctx.bound[opt]       = ret;
    case 'Connect'  : return ctx.connected[opt]   = ret;
  }
}
function deasync (fn) {
  return function () {
    var done=false, err, res;
    fn.apply(this, Array.prototype.slice.apply(arguments).concat(cb));
    while (!done) nn.block();
    if (err) throw err;
    return res;
    function cb (e, r) { err=e; res=r; done=true; };
  }
}
```

reqshark: > We're not using [ret] after invoking this.sync

yea `this.sync` is a deasync'd `function block()`:
```js
function block (nnMethod, ctx, opt, ret, done) {
  var throttle = setInterval(function(){
    if (ret !== undefined) {
      clearInterval(throttle);
      done();
    }
  }, 10);
  ret = ctx._protect(nn[nnMethod]( ctx.binding, opt ));
  switch (nnMethod) {
    case 'Bind'     : return ctx.bound[opt]       = ret;
    case 'Connect'  : return ctx.connected[opt]   = ret;
  }
}
```

i mean this line in deasync screws that idea up by unfortunately rewriting the function signature:
```js
fn.apply(this, Array.prototype.slice.apply(arguments).concat(cb));
```
to fix that we'll need to maintain the `arguments` object better. `ret` ends up not defined where it needs to be at least `undefined` inside what was `function block()`. I suppose passing `undefined` directly to `this.sync` will do the trick...

nickdesaulniers:
I've been thinking a lot about this.  Creating an async version of these functions in C++ and trying to "deasync" them to provide sync versions of those functions results in unnecessary complexity in the JS layer.  Contrast the patch's JS in lib/index.js, vs this:

```js
Socket.prototype.connect = function (addr, fn) {
  if (fn) {
    asyncConnect(this.binding, addr, function (eid) {
      // not sure if we need _protect around here?
      this.connected[addr] = eid;
    });
  } else {
    return this.connected[addr] = this._protect(nn.Connect( this.binding, addr ));
  }
};
```

nickdesaulniers:
updated my snippet

nickdesaulniers:
Further, JS already has faculties for "deasync": Promises + Generators.  Theoretically, we could provide only async versions of bind and connect, and let devs use Promises + Generators if they so desire.  But since we didn't rip out the synchronous C++ implementations of bind and connect, why not expose both sync and async versions?

reqshark:
Yea it would be simpler to expose two instead of trying to make one really complicated one.

PR #116, opened 2015-09-10
check EINTR after nn_close Fixes #115

### LINKED ISSUE #115: nn_close needs to check errno for EINTR
from the man page: http://nanomsg.org/v0.6/nn_close.3.html

we're not checking the return code then errno: https://github.com/nickdesaulniers/node-nanomsg/blob/master/src/node_nanomsg.cc#L35

nickdesaulniers:
let's see how this does @reqshark


reqshark:
this is good stuff, I'll merge now. and we can improve this more by firing a JS callback once the connection succeeds

PR #117, opened 2015-09-11
add optional async callback param to connect()

reqshark:
review @nickdesaulniers ?

if we can add some optional async callbacks to nanomsg operations like this, then i think we can guarantee ordering and free us from excessive reliance on `setTimeout` in our tests.


reqshark:
i think it makes sense to move this and the other async stuff we want over to #114, so actually dont merge that yet

PR #122, opened 2015-10-12
Test

nickdesaulniers:
very sloppy, will need more work.  Just want to see if it still hangs on travis or not.

### DIFF
--- a/test/chan.js
+++ b/test/chan.js
@@ -13,58 +13,59 @@ test('adding and removing subscription channels', function (t) {
   });
 
   pub.bind('inproc://filter');
-  sub.connect('inproc://filter');
+  sub.connect('inproc://filter', function () {
 
-  //the main array we'll use to test channel registration and removal
-  var msgs = ['hello world','hello world','hello world','bar world','foo world'];
-  var msgqty = -1; //starting here so first msg counted can be zero
-  var sent = 0;
+    //the main array we'll use to test channel registration and removal
+    var msgs = ['hello world','hello world','hello world','bar world','foo world'];
+    var msgqty = -1; //starting here so first msg counted can be zero
+    var sent = 0;
 
-  sub.on('data', function (buf) {
+    sub.on('data', function (buf) {
 
-    var msg = String(buf);
+      var msg = String(buf);
 
-    //increment msgqty when a msg is received (happening about every 100ms)
-    //we'll call finish() on the 5th msg received
-    switch(++msgqty){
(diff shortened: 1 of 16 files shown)


nickdesaulniers:
so the test failures are interesting. basically the specifics of the error message from nn.Err.  I get the two different ones based on whether I run `npm t` vs `make check`.  I wonder what the difference is, and whether it matters?

nickdesaulniers:
actually, the one version of node that passed has a curios failed assertion

test/chan.js
```diff
@@ -196,49 +197,50 @@ test('multi-topic registration followed by calls to rmchan on all but one stops
...
+        sub.rmchan('foo')
+      } else if(msgs === 2) {
+        t.equal(msg, 'bar world');
+        sub.rmchan('bar');
+      } else if (msgs === 3) {
+        t.equal(msg, 'hi world');
+        sub.rmchan('hi');
+      } else if (msgs === 4) {
+        t.equal(msg, 'hello world');
+      } else if (msgs === 5) {
```
reqshark: lets make that one a `switch(msg) { case n: // ...`

nickdesaulniers:
tests are hung, I don't having connect be async makes a difference.

reqshark:
@nickdesaulniers, The async worker threads are helpful but we're skirting around the real issue here, which is [endpoint identifiers](http://nanomsg.org/v0.6/nn_connect.3.html), see e.g. RETURN VALUE 

And while we're at it let's just quickly do a sanity check on your earlier question re: executing `npm i` vs. `make`, which still just executes `npm i` 

> I wonder what the difference is, and whether it matters?

the difference between those operations is zero.

we'll need to deal with the fact nanomsg connect() calls over TCP give us an *endpoint identifier*, which is not the same as an established connection state. Similarities around the ordering observed between those two states is coincidental, (although we really need a lot of latency to experience actual msg loss).

nickdesaulniers:
> we'll need to deal with the fact nanomsg connect() calls over TCP give us an endpoint identifier, which is not the same as an established connection state. Similarities around the ordering observed between those two states is coincidental

so when using TCP as the transport, calling `nn_connect` is not a guarantee of a connection?

reqshark:
correctamundo

reqshark:
well, no i mean eventually it is going to connect.. it has a nice backoff and retry algorithm if it gets really congested or takes a really long time for the bind address to come up..  basically one of the greatest features of nanomsg IMO.. who cares when clients or servers connect, disconnect, reconnect :palm_tree:

Issue #123, opened 2015-10-14
the build error: socket connection latencies and endpoint identifiers


reqshark:
[![Build Status](https://travis-ci.org/nickdesaulniers/node-nanomsg.svg?branch=master)](https://travis-ci.org/nickdesaulniers/node-nanomsg)

@nickdesaulniers likes pushing commits to master ~~and going to bed before the test finishes its run :tada:~~ 

whatever, at this point we've explored typical JavaScript perspective approaches to an idiosyncratic build error that showed up right after publishing `v2.1.2`

since then I've thought a lot about this problem. 

We agree that our socket's `connect()` return value is an endpoint identifier, so it doesn't matter whether the result is given to JavaScript from an async callback or as a blocking/synchronous return

from #122:

> when using TCP as the transport, calling nn_connect is not a guarantee of a connection

it's a *slow joiner syndrome*, an effect of the way the publish/subscriber pattern was designed. The zeromq guidebook breaks it down in the section: [`Getting The Message Out`](http://zguide.zeromq.org/page:all#Getting-the-Message-Out)

Since nanomsg inherited its PUB/SUB architecture from ZeroMQ's PUB/SUB pattern it's worth reading through the official guidebook (btw, O'Reilly published it.. and it's one of the best books written about software IMO.)

The relevant part for our travis test use case:

> There is one more important thing to know about PUB-SUB sockets: you do not know precisely when a subscriber starts to get messages. Even if you start a subscriber, wait a while, and then start the publisher, the subscriber will always miss the first messages that the publisher sends. This is because as the subscriber connects to the publisher (something that takes a small but non-zero time), the publisher may already be sending messages out.

That's all good, but there are two specific things I look forward to now:

1. Let's figure out a meaningful way to fix *slow joiner syndrome* in our tests <strong>without</strong> using `setTimeout`
2. Improve test result output and exit code

on that note, @nickdesaulniers, seems like the change from the last commit caused this:
```go
create loopback device with one socket
    ✓ should be equal
    ✓ error was thrown when device collapsed:Error: lib err: 156384765 Inappropriate ioctl for device
  tests 2
  pass  2
  throw exception when sending on socket after term() called
    ✓ library termination error thrown on send after term
  tests 1
  pass  1
0 passing (5.5s)
1003 failing
make: *** [check] Error 235
The command "make check" exited with 2.
Done. Your build exited with 1.
```

nickdesaulniers:
[...]

Issue #119, opened 2015-09-14
bad form in tests/standalone


reqshark:
`term()` test:

https://github.com/nickdesaulniers/node-nanomsg/blob/master/test/standalone/term.js

not sure what error gets tested. we're just seding or awking there without calling bind or connect.

PR #124, opened 2015-10-14
fix term err test: only test for an err

reqshark:
should fix the [Build Status](https://travis-ci.org/nickdesaulniers/node-nanomsg)

follows up #119, fixes a different type of build error than what #123 is about, to be fair, i edited my cheap shot taken in that issue too

I've noticed that sometimes testing for `err.message` equality gets lost on older versions of node on linux

### DIFF
--- a/test/standalone/term.js
+++ b/test/standalone/term.js
@@ -10,10 +10,8 @@ test('throw exception when sending on socket after term() called', function (t) 
   sock.bind('tcp://127.0.0.1:9999')
 
   sock.on('error', function (err) {
-    t.equals(err.message,
-      'Nanomsg library was terminated',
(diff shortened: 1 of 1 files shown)


nickdesaulniers:
That fancy build error button provided by travis links to the latest PR, regardless of whether or not its been merged. Notice how now it's linking to a run of this PR

reqshark:
i'm seeing it.

i just went to our settings interface on travis, it's basically empty. this is probably not the type of config we normally do in `.travis.yml`

nickdesaulniers:
passing test: https://travis-ci.org/nickdesaulniers/node-nanomsg/builds/85381246
merging

PR #125, opened 2015-10-17
add test and example using websocket transport

reqshark:
to checkout the `ws` transport example:
```bash
cd examples/ws
node server
```
then open your browser to `http://localhost:3000`

@nickdesaulniers review?

### DIFF
--- a/examples/ws/index.html
+++ b/examples/ws/index.html
@@ -1,0 +1,42 @@
+<!DOCTYPE html>
+<html>
+  <head></head>
+  <body>
+    <p>nanomsg ws test:&nbsp;<span></span></p>
+    <input type='text'>
+    <button>send</button>
+    <script>
+      var ws = new WebSocket('ws://127.0.0.1:7789', [
+      // see the rfc on sp websocket mapping:
+      // raw.githubusercontent.com/nanomsg/nanomsg/master/rfc/sp-websocket-mapping-01.txt
+        'pair.sp.nanomsg.org'
+      ]);
+      ws.onopen = function open() { ws.send('Ping'); }
+      ws.onerror = function error(err) { console.log(err); }
+      ws.onmessage = function message(e) {
+        var reader = new FileReader();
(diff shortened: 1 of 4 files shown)


nickdesaulniers:
does nanomsg understand WS:// natively? is it in our branch of libnanomsg 0.6?  Very cool!

reqshark:
yep, and if you want something more resilient to browser refreshes than pair, which I think only can handle one connection, then try the other socket types like pub/sub.

nickdesaulniers:
interesting, this doesn't work in Gecko, but does Blink/Safari. Anyways, I'll file a bug in Gecko and merge this. neat example

nickdesaulniers:
arg, I can't remember how to change it so that I was the commiter of the patch. Now we have a merge commit

reqshark:
@nickdesaulniers it's all here: https://github.com/jonathanong/git-style-guide/blob/master/README.md

reqshark:
btw, here's my workflow for doing git like a boss without merge commits:

1. `git checkout feature-branch`
2. `git rebase master`
3. if there is more than one commit, i squash it, so: `git rebase -i master`
4. `git commit --amend` i do that to add a bullet point in the commit saying `• PR: #555`
5. `git push -f` force push back to the branch
6. `git checkout master`
7. `git merge feature-branch`
8. `git push`

... woah i didn't realize there were so many steps involved! 

well anyway, that's how I personally do it. :boat:

reqshark:
> this doesn't work in Gecko

weird, I copied that FileReader stuff from Mozilla's website. 

nanomsg sends back a blob but you might be able to decode it differently

nickdesaulniers:
> weird, I copied that FileReader stuff from Mozilla's website. 

it's not the FileReader, but the websockets: https://bugzilla.mozilla.org/show_bug.cgi?id=1216237

reqshark:
@nickdesaulniers this might be related to https://github.com/nanomsg/nanomsg/issues/489

PR #126, opened 2015-10-20
add helper function to skip travis timeouts

reqshark:
For higher latency tcp connect() operations, we know it's possible for a socket to start publishing messages before a subscriber connection has been established, a.k.a. slow joiner syndrome

the helper function here avoids that issue in resulting travis timeouts and erroneous build errors 

```js
/*
 * skip function:
 *  • calls t.skip(msg), so the test passes after skipping the timed out test
 *  • closes any sockets given in an Array, the sock param
 */

function skip (t, subject, sock) {
  var msg = 'TEST TIMEOUT WARNING: no output after 10 seconds, skipping ';
  var skips = (t._plan - t.assertCount); // skips: pending number of tests
  var l = sock.length;

  while (l--) sock[l].close();
  while (skips--) t.skip(msg + subject + ' test');
}
```
implemented here in the streams and ws tests using a `setTimeout()` of 10 seconds and passing any outstanding (open) sockets to `skip()`, so those might be closed at the expiration of the timeout

during a successful travis run, we'll `clearTimeout()` along with the customary `socket.close()` cleanup

### DIFF
--- a/test/streams.js
+++ b/test/streams.js
@@ -2,7 +2,7 @@ var nano = require('..');
 var test = require('tape');
 
 test('pipe a thousand msgs between incompatible socket types', function(t){
-
+  t.plan(1);
   var sent = 0, recv = 0;
 
   var pub = nano.socket('pub', { tcpnodelay: true });
@@ -20,19 +20,34 @@ test('pipe a thousand msgs between incompatible socket types', function(t){
 
   sub.pipe(push);
   pull.on('data', pullsocket);
-
-  setTimeout(function(){
-    while(sent++ < 1001) pub.send('hello from nanomsg pub socket!');
-  }, 200);
-
   function pullsocket(msg){
     if(recv++ > 999){
-      t.equal( msg, 'hello from nanomsg pub socket!', 'piped a pub/pull combo');
       pub.close();
       push.close();
       pull.close();
       sub.close();
-      t.end();
+      t.equal( msg, 'hello from nanomsg pub socket!', 'piped a pub/pull combo');
+      clearTimeout(timeo);
     }
   }
+
+  var timeo = setTimeout(skip, 10000, t, 'streams', [pub, sub, push, pull]);
+
+  // start sending
+  while(sent++ < 1001) pub.send('hello from nanomsg pub socket!');
 });
+
+/*
+ * skip function:
+ *  • calls t.skip(msg), so the test passes after skipping the timed out test
+ *  • closes any sockets given in an Array, the sock param
+ */
+
+function skip (t, subject, sock) {
+  var msg = 'TEST TIMEOUT WARNING: no output after 10 seconds, skipping ';
+  var skips = (t._plan - t.assertCount); // skips: pending number of tests
+  var l = sock.length;
+
+  while (l--) sock[l].close();
+  while (skips--) t.skip(msg + subject + ' test');
+}
--- a/test/ws.js
+++ b/test/ws.js
@@ -1,6 +1,11 @@
-// This test suite is a duplicate of inproc.js, but using the ws transport.
+/* a transform stream implementation for piping msgs later in the test */
+function thr (fn) {
+  this._transform = fn;
+  require('stream').Transform.call(this);
+}
+require('util').inherits(thr, require('stream').Transform);
 
-var nano = require('../');
+var nano = require('..');
 var test = require('tape');
 
 test('ws socket pub sub', function (t) {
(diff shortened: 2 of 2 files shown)


reqshark:
oh and some revision to the ws bus test:
* removed the immediately invoked function used to capture for-loop iteration count
* pipe the bus sockets to a transform stream, replacing `.on('data'` callback

test/ws.js
```diff
@@ -208,18 +219,36 @@ test('ws multiple socket pub sub', function (t) {
...
+ */
+
+function skip (t, subject, sock) {
+  var msg = 'TEST TIMEOUT WARNING: no output after 10 seconds, skipping ';
+  var skips = (t._plan - t.assertCount); // skips: pending number of tests
+  var l = sock.length;
+
+  while (l--) sock[l].close();
+  while (skips--) t.skip(msg + subject + ' test');
+}
```
nickdesaulniers: do we need this functionality in the ws:// tests? I mean, I guess ws:// is usually built on top of tcp://, but I didn't think we've seen failures here?

test/streams.js
```diff
@@ -20,19 +20,34 @@ test('pipe a thousand msgs between incompatible socket types', function(t){
...
+ */
+
+function skip (t, subject, sock) {
+  var msg = 'TEST TIMEOUT WARNING: no output after 10 seconds, skipping ';
+  var skips = (t._plan - t.assertCount); // skips: pending number of tests
+  var l = sock.length;
+
+  while (l--) sock[l].close();
+  while (skips--) t.skip(msg + subject + ' test');
+}
```
nickdesaulniers: I'm super hesitant about disabling failing tests, or skipping them.  Some thoughts:
1. if we change the protocol from tcp:// to anything else, does this clear up? if so, then looks like slow joiner.
2. it could actually be covering up a bug or even a race condition in the newly added streams functionality.

test/ws.js
reqshark: yea those are even slower than TCP, I restarted the builds each time, see: 

* <strong>204</strong>: i just cancelled it after i saw it stalled around 6 mins: https://travis-ci.org/nickdesaulniers/node-nanomsg/builds/85972885

* <strong>205</strong>: failed for timeout: https://travis-ci.org/nickdesaulniers/node-nanomsg/builds/85973986
* <strong>206</strong>: same thing: https://travis-ci.org/nickdesaulniers/node-nanomsg/builds/85975846

test/streams.js
reqshark: Ya I figured you wouldn't be too into this one. `skip` is based on a timeout, not really a failing test

reqshark:
so @nickdesaulniers should we close this PR then? we could revisit it if timeouts become an issue later

nickdesaulniers:
let's keep it open and try to find the underlying issue with streams in the meantime.  We might have to resort to skipping timeouts, but let's keep it as a last resort.

reqshark:
look at the current build, we're still getting timeouts like crazy https://travis-ci.org/nickdesaulniers/node-nanomsg

nickdesaulniers:
I was thinking more about this last night.  I was thinking about how when I added console.log's to the counter variable in the timeout tests, we had a hard time replicating the timeouts on travis. I suspect that when we timeout, we're probably off by 1 for the count or something, but logging does something to the event loop that makes it hard to catch.

So the idea I had yesterday was from a college embedded programming project.  In embedded software, it's common to make use of a "watchdog timer."  Basically, at startup you initialize the watchdog, then as operations are going smoothly, you periodically "pet" the watchdog which resets the countdown.  If for whatever reason, the watchdog is not pet within a certain deadline, the timer reboots the hardware.  Kind of a dead man's switch.

So the code you wrote is useful on it's own as a watchdog library. A quick survey of NPM turned up https://www.npmjs.com/package/watchout which could also be used.  I think we should add watchdog timers to the tests that are timing out, that way, we can explicitly fail the tests rather than wait 10mins for timeouts.  We could even log the counter after this occurred, and that might give us some hints as to what's going wrong.  I suspect we're off by one.  If the that's the case, it will be more work to figure out why.  What do you think @reqshark ?

reqshark:
yea a watchdog timer is definitely in the spirit of things here. 

The snag detail to watch out for: sockets opened at the start of the test still need to be closed, and any unclosed sockets will prevent the test from completion.. so that would cause an infinite timeout.

`skip()` is doing some of the same stuff in that npm package `watchout`, but also accepts an array of sockets that we'll need to `close()` for everything to move forward smoothly.
```js
var timeo = setTimeout(skip, duration, t, msg, [sock1, sock2, sock3]);
```

in event of a successful run, we'll need to manually close the interval, preventing skip() from firing:

```js
clearTimeout(timeo);
```

reqshark:
there are better ways to fix the timeout than this PR

PR #129, opened 2015-10-26
Bug105

nickdesaulniers:
review? @reqshark 

can you try to test this on OSX? I'm running into an assertion failure on test/bind.js.  Looking into it now.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -162,7 +162,7 @@ Socket.prototype._protectArray = function (ret, unwind) {
 Socket.prototype._send = function (msg, flags) {
     if (this.closed) return;
 
-    if(this.type == 'surveyor') {
+    if(this.type === 'surveyor' || this.type === 'rep') {
         this._startPollReceive();
     }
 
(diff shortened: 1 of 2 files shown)


nickdesaulniers:
so the failure I was seeing was that there's a file left behind from test runs called some_address, which you can see in test/bind.js.  It looks like if the tests fail, then the file is left behind in a state that makes subsequent tests runs fail assertions.  I get the feeling `flock (2)` is called on this file, and it's never unlocked.  We should unconditionally remove this file after test runs.

reqshark:
I'll take a look... btw nice websocket timeout

nickdesaulniers:
> btw nice websocket timeout

arg!!!

nickdesaulniers:
also, do you think we should remove iojs from travis tests?

reqshark:
I think iojs is a legit version of node, but we could remove it if you want. I use node 4 now so I dont really care about testing it anymore

nickdesaulniers:
while the WS stuff did time out, the test I added passed: `multiple req sockets should not throw` https://travis-ci.org/nickdesaulniers/node-nanomsg/jobs/87511766.  That's not indicative of me not breaking anything though...

reqshark:
that's why skip() would have fired and it would have been appropriate

lib/index.js
```diff
@@ -162,7 +162,7 @@ Socket.prototype._protectArray = function (ret, unwind) {
 Socket.prototype._send = function (msg, flags) {
     if (this.closed) return;
 
-    if(this.type == 'surveyor') {
+    if(this.type == 'surveyor' || this.type == 'rep') {
```
reqshark: i think we should use trip equals

lib/index.js
```diff
@@ -186,6 +186,8 @@ Socket.prototype._receive = function () {
             this.emit('survey-timeout');
             return null;
         }
+    } else if(this.type == 'rep') {
```
reqshark: ditto

reqshark:
looks like it's passing on osx 10.11

PR #134, opened 2015-10-27
Support dynamic linking of libnanomsg

sirudog:
# NOTE: 
At the moment this option is strictly for linux. The main reason is that I could only test it on linux, although I see pkg-config can be installed on osx and even on windows (according to their wiki page).  
I mention this in the changed README.md.
I know @reqshark uses osx, so maybe you can test this, then the condition can be less strict with a simple change once it works there, too.


nickdesaulniers:
this doesn't break anything in OSX, or my linux distro.  Thanks for the PR!  Will squash+merge

nickdesaulniers:
merged, thank you!

PR #136, opened 2015-10-29
add watchdog to WS test that's timing out

nickdesaulniers:
let's see what else times out now, and I'll add subsequent patches.

### DIFF
--- a/test/ws.js
+++ b/test/ws.js
@@ -2,6 +2,7 @@
 
 var nano = require('../');
 var test = require('tape');
+var Watchout = require('watchout');
 
 test('ws socket pub sub', function (t) {
     t.plan(1);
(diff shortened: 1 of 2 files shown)


nickdesaulniers:
This is in the spirit of "test failures are preferable to test timeouts, but this is merely a band-aid until we figure out what's the root cause."

nickdesaulniers:
omg does the watchdog timer even work. wtf

nickdesaulniers:
or, maybe I added it to the wrong test. ugh doofus :anger:

test/ws.js
```diff
@@ -145,6 +160,8 @@ test('ws socket bus', function (t) {
 
         // Tally messages from other buses.
         bus.on('data', function (msg) {
+          watchdog.reset();
```
reqshark: why reset here?

reqshark:
haha ur comments are funny

test/ws.js
nickdesaulniers: for this library, `reset` is akin to petting the watchdog.  It does reset the internal timer.  If the timer ever falls to 0, :boom:

reqshark:
@nickdesaulniers, you don't like `skip()` in the other PR because it just skips the questionable test instead of running it. understandable.

So now what's the objective here? Instead of `skip()`, you basically want `retry()`? Is that right?

Issue #141, opened 2015-12-17
unexpected surveyor/respondent result


reqshark:
responses only show up once, and calling `survey()` again later eventually results in: 

```js
(node) warning: possible EventEmitter memory leak detected. 11 data listeners added.
Use emitter.setMaxListeners() to increase limit.
Trace
    at Socket.addListener (events.js:239:17)
    at Socket.Readable.on (_stream_readable.js:665:33)
    at Socket.survey (/home/reqshark/node_modules/nanomsg/lib/index.js:329:10)
```

reqshark:
until we fix this, here's a possible work-around:

```js
// heartbeat style survey every ~2 seconds

var nano  = require('nanomsg'), sur;
surveyor()
function surveyor(){
  sur = nano.socket('surveyor')
  sur.bind('tcp://0.0.0.0:12345')
  setTimeout(()=> sur.survey('knock knock', res), 2000); // enough time for TCP ESTABLISH
  function res (msg) {
    console.log('responses:', msg)
    sur.close()
    surveyor()
  }
}
```

reqshark:
confirmed to work. the reason i got EventEmitter leak stuff was because I was trying to call survey operations multiple times before `socket.bind(addr)` ever had a chance at TCP ESTABLISH 

so giving `bind()` a timeout and connecting the respondents first seems to resolve the issue and the workaround above is therefore unnecessary. I was going to close this issue, except for some slightly unexpected output. It looks like increasingly stacked answers given to the responses array

first if we connect a couple `a.js` and `b.js` respondent sockets:
```js
// a socket
var nano = require('nanomsg')
res = nano.socket('respondent')
res.connect('tcp://127.0.0.1:12345')
res.on('data', ()=> res.send('a is online'))
```

```js
// b socket
var nano = require('nanomsg')
res = nano.socket('respondent')
res.connect('tcp://127.0.0.1:12345')
res.on('data', ()=> res.send('b is online'))
```

then bind the survey socket:
```js
var nano = require('nanomsg')
sur = nano.socket('surveyor')
sur.bind('tcp://127.0.0.1:12345')

setTimeout(()=> sur.survey('sup', res), 100)
setTimeout(()=> sur.survey('sup', res), 200)
setTimeout(()=> sur.survey('sup', res), 300)
setTimeout(()=> sur.survey('sup', res), 400)

function res (msg) {
  console.log(msg
    .map(function(i){
      return String(i)
    }))
}
```

[...]

PR #142, opened 2015-12-30
Removing type specific APIs

nucleardreamer:
These type-specific APIs should be left out of the main lib. There should be a better section of examples in the readme or new JS examples to explain and document specific events (`survey-timeout` for example)


nickdesaulniers:
the corresponding tests need to be removed as well.

> There should be a better section of examples in the readme or new JS examples to explain and document specific events (survey-timeout for example)

Then add them in a PR.

Not sure if we want to keep this method or not. Will mull it over another day.

nucleardreamer:
Ahh, yes indeed I missed that! How about gathering all documentation of type specific APIs into one release? I definitely agree that is a good thing to have them in here, but it should be thorough and documented with some slick examples.

lib/index.js
```diff
@@ -312,26 +312,6 @@ Socket.prototype.setsockopt = function (level, option, value) {
...
-Socket.prototype.survey = function (buf, callback) {
-    if (!this.type === 'surveyor') {
-        throw new Error('Only surveyor sockets can survey.');
-    }
-    var responses = [];
-    function listener (buf) {
-        responses.push(buf);
-    }
-    this.once('survey-timeout', function () {
-        this.removeListener('message', listener);
```
reqshark: where are we setting the `message` event?

reqshark:
+1 :rocket: boom!

type-specific API oversteps into node-nanomsg user concern. 

this stuff was built around strong assumptions of specific user case. I would love to see it gone, as that would be a win for simplicity and predictability

moreover this is undocumented and not implemented correctly. that's very problematic and creates confusion, see #141. 

@nickdesaulniers, the `survey` method deleted here removes an `onmessage` event that was changed to `ondata` when we converted the recv mechanism to streams.

@nucleardreamer, please also remove the `emit` call on line 187 https://github.com/nickdesaulniers/node-nanomsg/blob/master/lib/index.js#L187

guys, i think it would be great to see a small test here based on the outcome of #141 

Thanks @nucleardreamer, this is much needed cleanup!

nickdesaulniers:
if @nucleardreamer removes the extra emit, this patch is good to go. Thanks! :+1: :baseball: :battery:

nucleardreamer:
Nice, Simple things FTW!

reqshark:
thanks! landed in b5533a3b3581f3836c86da664e8d37a5f2ecea00

PR #109, opened 2015-09-10
suppress compiler warnings on linux

reqshark:
### some gyp cleanup and re-org.  

here's a summary of the changes: 
• removed cflags for osx
• suppress compiler warnings on linux
• list collaborators in package.json
• update travis tests, covering node v0.10 up through latest (v4.0.0) 

####  why should we stop using cflags for darwin?
at this point, I've come to the realization that setting node gyp `cflags` for osx bindings and binary builds is a futile exercise. the cflags array (or equivalent string) is just ignored and basically broken. just try adding the correct `cflags` to quiet clang/xcode, or any other compile-time options for that matter and you'll see what i mean. therefore let's remove what we had configured for osx `cflags`. 

####  on the other hand, cflags work well on linux
node-nanomsg installs on linux will be nice and quiet from now on. This removes the following: 

```bash
In file included from ../deps/nanomsg/src/aio/usock.c:28:0:
../deps/nanomsg/src/aio/usock_posix.inc: In function ‘nn_usock_recv_raw’:
../deps/nanomsg/src/aio/usock_posix.inc:1134:21: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
                     *self->in.pfd = *((int*) CMSG_DATA (cmsg));
                     ^
../deps/nanomsg/src/aio/usock_posix.inc:1138:21: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
                     nn_closefd (*((int*) CMSG_DATA (cmsg)));
```
and
```bash
../deps/nanomsg/src/transports/utils/base64.c: In function ‘nn_base64_decode’:
../deps/nanomsg/src/transports/utils/base64.c:79:9: warning: array subscript has type ‘char’ [-Wchar-subscripts]
         ch = DECODEMAP [in [ii]];
```
and 
```bash
../deps/nanomsg/src/transports/ipc/bipc.c: In function ‘nn_bipc_start_listening’:
../deps/nanomsg/src/transports/ipc/bipc.c:387:9: warning: unused variable ‘opt’ [-Wunused-variable]
     int opt;
         ^
```
and 
```bash
In file included from ../deps/nanomsg/src/transports/tcpmux/../../utils/err.h:33:0,
                 from ../deps/nanomsg/src/transports/tcpmux/ctcpmux.c:37:
../deps/nanomsg/src/transports/tcpmux/ctcpmux.c: In function ‘nn_ctcpmux_create’:
[...]

PR #144, opened 2016-01-09
OTHER_CFLAGS: the long overdue OSX/node-gyp epiphany

reqshark:
finally figured out stuff related to #109 https://github.com/nodejs/node-gyp/issues/594

:nut_and_bolt: :date: :nut_and_bolt: :nut_and_bolt:

OK here's the trick to pass real `cflags=` to node-gyp compiler toolchain from OSX

npm installing node-nanomsg on OSX historically generates warnings, like these unused variables:

```go
In file included from ../deps/nanomsg/src/aio/poller.c:32:
../deps/nanomsg/src/aio/poller_kqueue.inc:66:9: warning: unused variable 'rc' [-Wunused-variable]
    int rc;
        ^
../deps/nanomsg/src/aio/poller_kqueue.inc:90:9: warning: unused variable 'fd' [-Wunused-variable]
    int fd = hndl->fd;
        ^
2 warnings generated.
```

we can add `-Wno-unused` to the arg list for evaluation on gyp targets by listing *`OTHER_CFLAGS` under `xcode_settings`*. 

### `cflags=` doesn't work outside `xcode_settings`
Only the super magically reserved word: *`OTHER_CFLAGS`* works, so `cflags` and `CFLAGS` won't cut it

also, when i was fooling around in our `.gyp` stuff this morning, i noticed:
* the `'OS=="MAC"` [condition in `binding.gyp`](https://github.com/nickdesaulniers/node-nanomsg/blob/master/binding.gyp#L16-L26) is useless (condition in 2nd target), let's get rid of it
* macro definition `NN_HAVE_GCC` isn't the best assumption for all OS, so moved to linux.gypi
* applied the magic in macosx.gypi to finally shut clang up! :boom: :bomb: :boom:

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -13,17 +13,6 @@
         ['use_system_libnanomsg=="false"', {
           'dependencies': [ 'deps/nanomsg.gyp:nanomsg', ],
         }],
-        ['OS=="mac"', {
-          'ldflags': [ '-L<(PRODUCT_DIR)' ],
-          'libraries': [ '-L<(PRODUCT_DIR)' ],
-          'xcode_settings': {
-            'OTHER_CPLUSPLUSFLAGS': [
-              '-fexceptions',
-              '-Wall',
-              '-Werror'
-            ]
-          }
-        }],
         ['OS=="linux" and use_system_libnanomsg=="true"', {
           'include_dirs+': [
             '<!@(pkg-config libnanomsg --cflags-only-I | sed s/-I//g)',
--- a/deps/common.gypi
+++ b/deps/common.gypi
@@ -1,7 +1,6 @@
 {
     # compiler settings to build the nanomsg library
     'defines': [
-        'NN_HAVE_GCC',
         'NN_HAVE_SOCKETPAIR',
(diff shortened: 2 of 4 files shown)


nickdesaulniers:
incredible! :snail: :fire: :church: :fire:

nickdesaulniers:
will merge later today

reqshark:
I know right?! Cheers man :beers:

yoshuawuyts:
:sparkles::sparkles::sparkles:

reqshark:
:mushroom: :cactus: :palm_tree: :octopus: :turtle:

d4tocchini:
thanks for the magic, @reqshark!

I'm late to this thread, but after wasting way too much on issues like this, I threw together a node-gyp wrapper, [node-gypcc](https://github.com/d4tocchini/node-gypcc), that bottles this sort of nasty and conforms to `$CC` cli like gcc/clang/emcc.  It's early PoF, but already proving super helpful, especially integrating with make/cmake builds, maybe it helps someone who finds themselves here, cheers

PR #149, opened 2016-02-23
Avoid copying received message buffers

kkoopa:
Don't Copy That Floppy

### DIFF
--- a/src/node_nanomsg.cc
+++ b/src/node_nanomsg.cc
@@ -108,6 +108,11 @@ NAN_METHOD(Send) {
   }
 }
 
+void fcb(char *data, void *hint) {
+  nn_freemsg(data);
+  (void) hint;
+}
+
 NAN_METHOD(Recv) {
   int s = Nan::To<int>(info[0]).FromJust();
   int flags = Nan::To<int>(info[1]).FromJust();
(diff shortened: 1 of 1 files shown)


reqshark:
woah :pager: avoiding buffer copy?! SICK! 

 Thanks @kkoopa! :palm_tree:

btw what happened to travis on this one? whatever, I got `894 passing (10.1s)` on linux/osx..

kkoopa:
Don't know what happened to Travis. I was wondering about it myself. Is it set up to test all PRs (the default)?

Out of curiosity, was there any speedup with the current (performance) tests?

reqshark:
Indeed with the speedup, even on the smallest bufs we're getting a nice 10% bump in throughput.. 

### before:
```
message size: 1 [B]
message count: 100000
throughput: 81940 [msg/s]
throughput: 0.656 [Mb/s]
```

### after:
```
message size: 1 [B]
message count: 100000
throughput: 98185 [msg/s]
throughput: 0.785 [Mb/s]
```

And as for travis,

> Is it set up to test all PRs (the default)?

It's always been that way.. although looking at the yaml it does say [`branches: only master`](https://github.com/nickdesaulniers/node-nanomsg/blob/master/.travis.yml#L10-L12)

reqshark:
yup, @travis-ci with the air ball! not even close to the rim...

hmm

kkoopa:
That's actually a 20 % increase. Now the remaining overhead is just 2 %. Anyway, nice that it did some good.

reqshark:
right, clearly a 20% increase in throughput there...

moreover, consistently improved performance across larger buffer ranges too

### before:
```sh
message size: 10 [B]
message count: 100000
throughput: 92122 [msg/s]
throughput: 7.37 [Mb/s]
```
### after:
```sh
message size: 10 [B]
message count: 100000
throughput: 117064 [msg/s]
throughput: 9.365 [Mb/s]
```

well done @kkoopa ! cheers! :beers:

yoshuawuyts:
sick!

PR #159, opened 2016-06-23
update system install option

### LINKED ISSUE #157: --use_system_libnanomsg=true is broken after the update to 0.9-beta
Because the pkg-config file include directory is now "/usr/local/include" and not "/usr/local/include/nanomsg" using the flag above causes the build to fail since it can't find <nn.h>

Fix would be most likely in binding.gyp to modify the include dir read from the pkg-config file. Also from 0.9-beta onwards the pkg-config file has changed from libnanomsg to nanomsg.

I'll probably get around to sending a PR for this but opening this issue in case anyone else can get to it before me.

Thanks.
Deepak.

reqshark:
fixes #157 for libnanomsg >= 1.0

BTW this `--use_system_libnanomsg` flag is supported on linux only

do we want to enable this for osx? could add that here while we're at it.

cc for review: @nickdesaulniers, @deepakprabhakara

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -15,10 +15,22 @@
         }],
         ['OS=="linux" and use_system_libnanomsg=="true"', {
           'include_dirs+': [
-            '<!@(pkg-config libnanomsg --cflags-only-I | sed s/-I//g)',
+            '<!@(pkg-config nanomsg --cflags-only-I | sed s/-I//g)/nanomsg  || echo "")',
+            '<!@(pkg-config libnanomsg --cflags | sed s/-I//g || echo "")',
           ],
           'libraries': [
-            '<!@(pkg-config libnanomsg --libs)',
+            '<!@(pkg-config nanomsg --libs || echo "")',
+            '<!@(pkg-config libnanomsg --libs || echo "")',
+          ],
+        }],
+        ['OS=="mac" and use_system_libnanomsg=="true"', {
+          'include_dirs+': [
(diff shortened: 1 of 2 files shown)


cehoffman:
OS X support would be very beneficial. As part of Homebrew, pkg-config works perfectly fine with nanomsg.

nickdesaulniers:
pkg-config has been hit or miss for me on OSX, but indeed it looks like the distribution of nanomsg works for me:

```
➜  align brew install nanomsg
==> Downloading https://homebrew.bintray.com/bottles/nanomsg-0.8-beta.el_capitan.bottle.tar.gz
######################################################################## 100.0%
==> Pouring nanomsg-0.8-beta.el_capitan.bottle.tar.gz
🍺  /usr/local/Cellar/nanomsg/0.8-beta: 106 files, 1.6M
➜  align pkg-config libnanomsg --libs
-L/usr/local/Cellar/nanomsg/0.8-beta/lib -lnanomsg
```

nickdesaulniers:
and I think pkg-config is not installed by default on OSX, so we should consider what happens when a user tries to use `--use_system_libnanomsg` but doesn't have pkg-config.

reqshark:
oh cool adding osx!

reqshark:
i think we can support older versions of libnanomsg and/or check the existence of different directories, including trying to use pkg-config by letting GYP fail gracefully like:


```gyp
'<!@(pkg-config nanomsg --cflags-only-I | sed s/-I//g)/nanomsg || echo ""'
```

PR #160, opened 2016-06-27
Make Send and Recv non blocking

alimousazy:
Make sure that we won't block on receive or send


reqshark:
Thanks @alimousazy! we'll cut a new node-nanomsg release with this change in the next day or two! :)

alimousazy:
Any plan to update libnanomsg dependency to 1.0 ?

reqshark:
ya that was updated here in our `3.2.0` release

PR #162, opened 2016-08-05
fix: not subscribed message cause stream end

a0000778:
sub.chan(['a']);
pub.send('b');//Socket._receive() return null cause Socket._startPollReceive() this.push(null)
pub.send('a');

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -176,7 +176,10 @@ Socket.prototype._send = function (msg, flags) {
 };
 
 Socket.prototype._receive = function () {
-  if (this.closed) return null;
+  if (this.closed){
+    this.push(null);
+    return;
+  }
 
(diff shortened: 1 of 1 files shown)


reqshark:
I like this change because now we're not necessarily pushing the result of `_receive()` onto the stream. 

I think this makes our recv operations safer overall and less prone to any stream errors

reqshark:
Thank you for the nice bug fix! 

@a0000778, can i ask your name to add you to node-nanomsg list of contributors?

a0000778:
I don't use my name on the internet.
Only add my GitHub ID to node-nanomsg list of contributors.

reqshark:
kk will do 😉

nickdesaulniers:
thank you!

Issue #151, opened 2016-03-02
memory leak


reqshark:
PollSendSocket leaks. At least valgrind thinks so.

``` bash
==4428== 2,112 bytes in 12 blocks are possibly lost in loss record 996 of 1,041
==4428==    at 0x4C2CC70: calloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==4428==    by 0x985CF78: PollSendSocket(Nan::FunctionCallbackInfo<v8::Value> const&) (in /home/reqshark/node_modules/nanomsg/build/Release/node_nanomsg.node)
==4428==    by 0x985C555: Nan::imp::FunctionCallbackWrapper(v8::FunctionCallbackInfo<v8::Value> const&) (in /home/reqshark/node_modules/nanomsg/build/Release/node_nanomsg.node)
==4428==    by 0x913C21: v8::internal::FunctionCallbackArguments::Call(void (*)(v8::FunctionCallbackInfo<v8::Value> const&)) (in /home/reqshark/.nvm/versions/node/v5.6.0/bin/node)
==4428==    by 0x93E100: v8::internal::MaybeHandle<v8::internal::Object> v8::internal::HandleApiCallHelper<false>(v8::internal::Isolate*, v8::internal::(anonymous namespace)::BuiltinArguments<(v8::internal::BuiltinExtraArguments)1>&) (in /home/reqshark/.nvm/versions/node/v5.6.0/bin/node)
==4428==    by 0x93E5F1: v8::internal::Builtin_HandleApiCall(int, v8::internal::Object**, v8::internal::Isolate*) (in /home/reqshark/.nvm/versions/node/v5.6.0/bin/node)
==4428==    by 0x1D5F9040839A: ???
==4428==    by 0x1D5F90644AB0: ???
==4428==    by 0x1D5F906448B7: ???
==4428==    by 0x1D5F90408296: ???
==4428==    by 0x1D5F9063F5EA: ???
==4428==    by 0x1D5F9066B954: ???
```

cc: @nickdesaulniers, @kkoopa, @tcr

kkoopa:
False alarm. It says possibly lost. Valgrind does not know about finalizers (and cannot reasonably be expected to). V8's garbage collector will clean it up when it sees fit to do so.

reqshark:
[...]

PR #167, opened 2016-11-26
add asan build support #151

nickdesaulniers:
build with `npm i --asan=true`.  This will use a build that links in
asan at runtime.

The odd split in .gyp files exists because we want to compile nanomsg
with asan instrumentation, but when we link our node_namomsg.node shared
object we want to remind gyp to link asan's runtime library.

This can likely also be done for Linux, but I haven't tested.

Also, we segfault off the bat with this.  I don't have time to look into
it, but at least it builds correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -420,3 +420,7 @@ WIP
 ## license
 
 MIT
+
+## memory leak hunting
+`npm i --asan=true`
+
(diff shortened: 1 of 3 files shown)


nickdesaulniers:
@reqshark

PR #169, opened 2016-12-18
Memleak

nickdesaulniers:
review? @reqshark @m-ohuchi @deepakprabhakara @k-keisuke @kkoopa


src/node_nanomsg.cc
```diff
@@ -191,35 +190,40 @@ NAN_METHOD(Err) {
...
 }
 
-typedef struct nanomsg_socket_s {
-  uv_poll_t poll_handle;
-  uv_os_sock_t sockfd;
-  Nan::Callback *callback;
-} nanomsg_socket_t;
+class NanoMsgPollCtx {
+  private:
+  public:
```
kkoopa: This is conceptually a `struct`. A `struct` would not require access modifiers.

src/node_nanomsg.cc
```diff
@@ -191,35 +190,40 @@ NAN_METHOD(Err) {
...
+    uv_poll_t poll_handle;
+    uv_os_sock_t sockfd;
+    Nan::Callback *callback;
+    NanoMsgPollCtx (int s, bool is_sender, Local<v8::Function> cb) {
+      callback = new Nan::Callback(cb);
```
kkoopa: Prefer using an initializer list.

src/node_nanomsg.cc
```diff
@@ -191,35 +190,40 @@ NAN_METHOD(Err) {
...
+      size_t siz = sizeof(uv_os_sock_t);
```
kkoopa: `sizeof` is a keyword (like `if`, and `for`) not a function, so add a space.

src/node_nanomsg.cc
```diff
@@ -191,35 +190,40 @@ NAN_METHOD(Err) {
...
+      nn_getsockopt(s, NN_SOL_SOCKET, is_sender ? NN_SNDFD : NN_RCVFD, &sockfd,
+          &siz);
+      poll_handle.data = this;
+    }
+    ~NanoMsgPollCtx () {
+      delete callback;
+    }
```
kkoopa: I'd make the `callback` member an actual `Nan::Callback` instead of a pointer. Then you would not need this destructor, (which incidentally disables automatic generation of move constructors, which might or might not be desirable).

src/node_nanomsg.cc
```diff
@@ -191,35 +190,40 @@ NAN_METHOD(Err) {
...
+};
 
-void NanomsgReadable(uv_poll_t *req, int status, int events) {
+static void NanomsgReadable(uv_poll_t *req, int status, int events) {
```
kkoopa: The `static` keyword is not needed for free functions, they are implicitly static. On the other hand, it is not wrong to add it.

src/node_nanomsg.cc
```diff
@@ -191,35 +190,40 @@ NAN_METHOD(Err) {
...
   Nan::HandleScope scope;
 
-  nanomsg_socket_t *context;
-  context = reinterpret_cast<nanomsg_socket_t *>(req);
+  NanoMsgPollCtx *context = reinterpret_cast<NanoMsgPollCtx*>(req->data);
```
kkoopa: prefer `static_cast`

src/node_nanomsg.cc
```diff
@@ -191,35 +190,40 @@ NAN_METHOD(Err) {
...
-  Nan::Callback *callback = new Nan::Callback(info[1].As<Function>());
-
-  nanomsg_socket_t *context;
-  size_t siz = sizeof(uv_os_sock_t);
-
-  context = reinterpret_cast<nanomsg_socket_t *>(calloc(1, sizeof *context));
-  context->poll_handle.data = context;
-  context->callback = callback;
-  nn_getsockopt(s, NN_SOL_SOCKET, NN_SNDFD, &context->sockfd, &siz);
+  bool is_sender = Nan::To<bool>(info[1]).FromJust();
```
kkoopa: I assume there is enough sanity checking on the JavaScript side to ensure this is not `Nothing`.

src/node_nanomsg.cc
```diff
@@ -229,30 +233,14 @@ NAN_METHOD(PollSendSocket) {
...
-  nn_getsockopt(s, NN_SOL_SOCKET, NN_RCVFD, &context->sockfd, &siz);
-
-  if (context->sockfd != 0) {
-    uv_poll_init_socket(uv_default_loop(), &context->poll_handle,
-                        context->sockfd);
-    uv_poll_start(&context->poll_handle, UV_READABLE, NanomsgReadable);
-    info.GetReturnValue().Set(WrapPointer(context, 8));
-  }
+static void close_cb(uv_handle_t *handle) {
+  NanoMsgPollCtx *context = reinterpret_cast<NanoMsgPollCtx*>(handle->data);
```
kkoopa: `static_cast`

src/node_nanomsg.cc
```diff
@@ -229,30 +233,14 @@ NAN_METHOD(PollSendSocket) {
...
+  delete context;
 }
 
 NAN_METHOD(PollStop) {
-  nanomsg_socket_t *context = UnwrapPointer<nanomsg_socket_t *>(info[0]);
-  int r = uv_poll_stop(&context->poll_handle);
-  info.GetReturnValue().Set(Nan::New<Number>(r));
+  NanoMsgPollCtx *context = UnwrapPointer<NanoMsgPollCtx*>(info[0]);
+  uv_close(reinterpret_cast<uv_handle_t*>(&context->poll_handle), close_cb);
```
kkoopa: `static_cast`

src/node_nanomsg.cc
nickdesaulniers: I tried that.  The `Nan::Callback` is marked no move/copy, so it must be heap allocated.

src/node_nanomsg.cc
nickdesaulniers: `sizeof` requires parens for evaluating types.  It does not need them for variables.

src/node_nanomsg.cc
nickdesaulniers: oops, wrong comment

src/node_nanomsg.cc
nickdesaulniers: I don't think so.  Any other compilation can forward declare `extern void NanomsgReadable ...` and the link against it (if `static` was removed).  Explicitly marking this function `static` restricts its linkage to this compilation unit.

src/node_nanomsg.cc
nickdesaulniers: Isn't `static_cast` for inheritance hierarchies?  `NanoMsgPollCtx` doesn't inherit from `void` so I don't see how that would work.

src/node_nanomsg.cc
nickdesaulniers: The two call sites pass literals.

src/node_nanomsg.cc
kkoopa: Why is that? I don't see instances of `NanoMsgPollCtx` being copied or moved. I only saw one being heap allocated and the pointer passed around. If it is because of no assignment operator on `Nan::Callback`, that would not be an issue when using intializer list.

src/node_nanomsg.cc
kkoopa: If restricting linkage is the point, they should be contained in an anonymous namespace. Marking functions static to restrict linkage is old-fashioned (maybe discouraged?) in C++.

src/node_nanomsg.cc
kkoopa: Just tried, it works fine.
```c++
class PollCtx {
    const Nan::Callback callback;
     PollCtx(const int s, const bool is_sender, const Local<v8::Function> cb): callback(cb) {}
}
```

src/node_nanomsg.cc
kkoopa: Now I realized you may have misunderstood what I was trying to say:
I claim that `sizeof ( uv_os_sock_t)` is better than `sizeof( uv_os_sock_t)`, since the rest of the code follows the convention of keywords having a space before their parameter list, e.g. `if (foo)` vs `if(foo)`.

src/node_nanomsg.cc
kkoopa: Yes, I merely noticed that everything was public and there were no methods, so calling it `struct` would have been more clear. It does not apply any longer, since you now changed its semantics to that of a `class`.

src/node_nanomsg.cc
kkoopa: To clarify some more, in C++, the type `void *` is different from any other `T *` in that `void *` is the top type (base type, e.g. `Object` in Java) of all (data) pointers. Any `T *` can be safely converted to `void *` and back. The reason this is quirky is that `void` itself is actually the bottom type, which cannot be converted to anything, nor can anything be converted to it.

To put it differently, `void *` is not a pointer to `void`, it is a pointer to an unknown type.

src/node_nanomsg.cc
nickdesaulniers: ah makes sense.  I'll try to get `clang-format` installed on this machine, then just defer to that.

src/node_nanomsg.cc
nickdesaulniers: was this changed in a new version of Nan or specific version of Node?  I had gotten a build error, but I will triple check now.

src/node_nanomsg.cc
nickdesaulniers: makes sense, maybe I should wrap all of this into an anonymous namespace then?  Will exporting the Nan functions still work then?

src/node_nanomsg.cc
nickdesaulniers: thanks for the explanation, I will try this again.

src/node_nanomsg.cc
kkoopa: No.
However, with C++11, you would get an error with `PollCtx(const int s, const bool is_sender, const Local<v8::Function> cb) : callback(Nan::Callback(cb)) {}` saying `Nan::Callback` is not movable.
With older C++ it would say not copyable.
That is because that line would move (or copy) construct `PollCtx::callback` from the (temporary) `Nan::Callback`.
Without initializer list, it would fail in either version due to `Nan::Callback` not being assignable.

nickdesaulniers:
```sh
../src/node_nanomsg.cc:213:9: error: call to deleted constructor of
      'const Nan::Callback'
        callback(Nan::Callback(cb)) {
        ^        ~~~~~~~~~~~~~~~~~
../node_modules/nan/nan.h:1427:33: note: 'Callback' has been explicitly marked deleted
      here
  NAN_DISALLOW_ASSIGN_COPY_MOVE(Callback)
                                ^
➜  node-nanomsg git:(memleak) ✗ node -v
v6.9.2
➜  node-nanomsg git:(memleak) ✗ cat package.json| grep nan
    "nan": "2.4"
```

src/node_nanomsg.cc
kkoopa: Hmm... I actually don't know. Node addons are dynamic libraries, so an export table containing function addresses needs to be built at link time. but I don't know if that affects anything.

src/node_nanomsg.cc
nickdesaulniers: ../src/node_nanomsg.cc:250:12: error: static_cast from 'uv_poll_t *' (aka 'uv_poll_s *')
      to 'uv_handle_t *' (aka 'uv_handle_s *'), which are not related by inheritance, is
      not allowed
  uv_close(static_cast<uv_handle_t*>(&context->poll_handle), close_cb);
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/node_nanomsg.cc
kkoopa: See above comment. It should say ` PollCtx() : callback(cb)`, not ` PollCtx() : callback(Nan::Callback(cb))`.

src/node_nanomsg.cc
kkoopa: Alright, seems this one was not a `void *`. Since `static_cast` does not work, `reinterpret_cast` is valid.

src/node_nanomsg.cc
nickdesaulniers: So I need to construct the `Nan::Callback` before constructing the `PollCtx`?

nickdesaulniers: oh I see what I need to do...

kkoopa: No, that would not work (since you could not set the member to the previously constructed callback.

Accessor lists work as such:
```c++
class Foo {
 public:
  Foo(int n) : n_(n) {}

 private:
    int n_;
};

class Bar {
 public:
  Bar(Foo f) : f_(f) {}

 private:
  Foo f_;
};
```
In `Bar(int n) : f_(n) {}`, read the `f_(n)` part as `Foo f_(n)`, e.g. same as in
```c++
int main() {
  int n = 4;
  Foo f_(n);

  return 0;
}

File: src/node_nanomsg.cc [hunk 10]
```diff
@@ -191,68 +190,61 @@ NAN_METHOD(Err) {
...
-
-  if (context->sockfd != 0) {
-    uv_poll_init_socket(uv_default_loop(), &context->poll_handle,
-                        context->sockfd);
-    uv_poll_start(&context->poll_handle, UV_READABLE, NanomsgReadable);
-    info.GetReturnValue().Set(WrapPointer(context, 8));
-  }
+  bool is_sender = Nan::To<bool>(info[1]).FromJust();
+  Local<v8::Function> cb = info[2].As<Function>();
+  PollCtx *context = new PollCtx(s, is_sender, cb);
```
nickdesaulniers: I think we can add some `const` here.

src/node_nanomsg.cc
```diff
@@ -191,68 +190,61 @@ NAN_METHOD(Err) {
...
-
-  if (context->sockfd != 0) {
-    uv_poll_init_socket(uv_default_loop(), &context->poll_handle,
-                        context->sockfd);
-    uv_poll_start(&context->poll_handle, UV_READABLE, NanomsgReadable);
-    info.GetReturnValue().Set(WrapPointer(context, 8));
-  }
+static void close_cb(uv_handle_t *handle) {
+  const PollCtx* const context = static_cast<PollCtx*>(handle->data);
+  delete context;
```
nickdesaulniers: probably want to assign handle->data null

src/node_pointer.h
```diff
@@ -4,63 +4,29 @@
...
-/**
- * Templated version of UnwrapPointer that does a reinterpret_cast() on the
- * pointer before returning it.
+ * Unwraps Buffer instance "buffer" to a PollCtx with the offset specified.
  */
 
-template <typename Type>
-inline static Type UnwrapPointer(v8::Local<v8::Value> buffer) {
-  return reinterpret_cast<Type>(UnwrapPointer(buffer));
+PollCtx* UnwrapPointer(v8::Local<v8::Value> buffer,
```
nickdesaulniers: I think I'd like to make these static methods of PollCtx.

nickdesaulniers:
hmm...tests are failing only remotely after moving PollCtx to its own compilation unit.  I wonder if I messed up the UnwrapPointer template?

nickdesaulniers:
the failed tests seem to always be the shutdown tests.

nickdesaulniers:
unless there are guarantees that a sub socket subscribed to multiple pub sockets is guaranteed to receive sends in order over tcp sockets (even during slowdowns on travis), then I don't think the shutdown test's clean up code is safe and would explain why it's failing. cc @reqshark @gdamore

m-ohuchi:
Hi @nickdesaulniers  I'm watching your changes. 
And I found valgrind reported the invalid memory access as follows:

    ==15742== Invalid write of size 4
    ==15742==    at 0x7B20501: close_cb(uv_handle_s*) (node_nanomsg.cc:204)
    ==15742==    by 0x8F039B1: uv__finish_close (core.c:272)
    ==15742==    by 0x8F039B1: uv__run_closing_handles (core.c:286)
    ==15742==    by 0x8F039B1: uv_run (core.c:356)
    ==15742==    by 0x8C8829D: node::Start(int, char**) (in     /core/bin/node)
    ==15742==    by 0x83A4964: main (in     /core/bin/node)
    ==15742==  Address 0x466c4d0 is 8 bytes inside a block of size 92 free'd
    ==15742==    at 0x4026703: operator delete(void*) (vg_replace_malloc.c:575)
    ==15742==    by 0x7B20500: close_cb(uv_handle_s*) (node_nanomsg.cc:203)
    ==15742==    by 0x8F039B1: uv__finish_close (core.c:272)
    ==15742==    by 0x8F039B1: uv__run_closing_handles (core.c:286)
    ==15742==    by 0x8F039B1: uv_run (core.c:356)
    ==15742==    by 0x8C8829D: node::Start(int, char**) (in     /core/bin/node)
    ==15742==    by 0x83A4964: main (in     /core/bin/node)
    ==15742==  Block was alloc'd at
    ==15742==    at 0x402768F: operator new(unsigned int) (vg_replace_malloc.c:327)
    ==15742==    by 0x7B20642: PollSocket(Nan::FunctionCallbackInfo<v8::Value> const&) (node_nanomsg.cc:197)
    ==15742==    by 0x7B20463: Nan::imp::FunctionCallbackWrapper(v8::FunctionCallbackInfo<v8::Value> const&) (nan_callbacks_12_inl.h:174)
    ==15742==    by 0x854CB66: v8::internal::FunctionCallbackArguments::Call(void (*)(v8::FunctionCallbackInfo<v8::Value> const&)) (in     /core/bin/node)
    ==15742==    by 0x85AC9F0: v8::internal::MaybeHandle<v8::internal::Object> v8::internal::(anonymous namespace)::HandleApiCallHelper<false>(v8::internal::Isolate*, v8::internal::(anonymous namespace)::BuiltinArguments<(v8::internal::BuiltinExtraArguments)1>) (in     /core/bin/node)
    ==15742==    by 0x85ACFBF: v8::internal::Builtin_HandleApiCall(int, v8::internal::Object**, v8::internal::Isolate*) (in     /core/bin/node)

The line node_nanomsg.cc:204 is following:
    handle->data = NULL; 

Please update this if you know how to fix. 
I'm examining the solution that we can simply disable this line or need to do something instead.

Thanks,

nickdesaulniers:
@ m-ohuchi, great catch! I'm glad your checking.  I'll revert that commit.

nickdesaulniers:
ok now I'm quite happy with this PR.  Ready for rereview.

src/poll_ctx.cc
```diff
@@ -0,0 +1,52 @@
...
+
+Local<Value> PollCtx::WrapPointer (void* ptr, size_t length) {
+   return Nan::NewBuffer(static_cast<char *>(ptr), length, wrap_pointer_cb, 0)
+     .ToLocalChecked();
+}
+
+PollCtx* PollCtx::UnwrapPointer (v8::Local<v8::Value> buffer, const int64_t
+    offset) {
+  return reinterpret_cast<PollCtx*>(node::Buffer::HasInstance(buffer) ?
+    node::Buffer::Data(buffer.As<v8::Object>()) + offset : 0);
```
nickdesaulniers: idk how I feel about this `offset`.  We don't use it, I think int64_t is not even the right type we want, and the pointer arithmetic is rather dangerous.  Since `WrapPointer` always uses a hardcoded value of `0`, I think I'll just remove this.

src/poll_ctx.cc
```diff
@@ -0,0 +1,52 @@
...
+}
+
+void PollCtx::invoke_callback (const int events) const {
+  Nan::HandleScope scope;
+  Local<Value> argv[] = { Nan::New<Number>(events) };
+  callback.Call(1, argv);
+}
+
+// Called when the "pointer" is garbage collected.
+static void wrap_pointer_cb(char * /* data */, void * /* hint */) {}
```
nickdesaulniers: this is dumb, can we remove it?

nickdesaulniers: no: https://github.com/nodejs/nan/blob/master/doc/buffers.md#nannewbuffer

"When a node::Buffer instance is garbage collected **and a FreeCallback has not been specified**, data will be disposed of via a call to free()."

src/node_nanomsg.cc
```diff
@@ -191,68 +184,22 @@ NAN_METHOD(Err) {
...
 }
 
 NAN_METHOD(PollStop) {
-  nanomsg_socket_t *context = UnwrapPointer<nanomsg_socket_t *>(info[0]);
-  int r = uv_poll_stop(&context->poll_handle);
-  info.GetReturnValue().Set(Nan::New<Number>(r));
+  PollCtx* const context = PollCtx::UnwrapPointer(info[0]);
```
nickdesaulniers: we should probably check this is not null

src/poll_ctx.cc
kkoopa: However, I think `node::Buffer::Data` (incorrectly) returns a `char *`, so it might require two `static_cast`s, first to get a `void *`, then to get a `PollCtx *`.

gdamore:
There are no such guarantees.  While TCP does have an ordering guarantee, nanomsg does not, and never will.  There are numerous ways that messages can be delivered out of order in nanomsg.

PR #174, opened 2017-03-10
add dontwait option

### LINKED ISSUE #173: NN_DONTWAIT triggers JS stream/event error for PUSH ipc:// under msg load
protocol might not matter, but some cause and effect observed in #172, reported by @dremekie: 

```
events.js:160
      throw er; // Unhandled 'error' event
      ^

Error: Resource temporarily unavailable
    at Socket._protect (/Users/dwayne/Documents/projects/test/node_modules/nanomsg/lib/index.js:144:28)
    at Socket._send (/Users/dwayne/Documents/projects/test/node_modules/nanomsg/lib/index.js:173:17)
    at Socket.flush (/Users/dwayne/Documents/projects/test/node_modules/nanomsg/lib/index.js:273:14)
    at Socket.<anonymous> (/Users/dwayne/Documents/projects/test/node_modules/nanomsg/lib/index.js:204:30)
```

Resource temporarily unavailable is associated with `EAGAIN` or `EWOULDBLOCK` -[`errno` man page](http://man7.org/linux/man-pages/man3/errno.3.html)

`NN_DONTWAIT` puts the socket into non-blocking I/O mode, which may lead to either an `EAGAIN` or `EWOULDBLOCK` when the send buffer is completely full. 

from my `send` manual page:
```bash
If no messages space is available at the socket to hold the message to be transmitted,
then send() normally blocks, unless the socket has been placed in non-blocking I/O mode
```

A PUSH socket's send buffer could max out when calls to `send()` accumulate without a drain from PULL socket's `recv()`. In cases like these, if the OS suggests `EAGAIN`, error: try again... an ideal retry would wait some time to allow PULL socket activity against the filled send buffer.

In Node.JS, users should listen for this error on a PUSH stream like `push.on('error', function(e)` and make sure to set a reasonable timeout before retrying that pipe.

In the meantime, 

@nickdesaulniers, we can:
* expose `NN_DONTWAIT` flag to the user as another generic piece of the `sockopt()` api, or
* don't set the send flag for PUSH (or problem could also be send flag on`ipc://`)

Not setting the flag seems to fix `Resource temporarily unavailable` in #172

reqshark:
fixes #173 

the NN_DONTWAIT flag is: 
* zero (for blocking) or
* one (for async I/O)

We can set the option (including override PUSH default) by doing:

```js
var push = require('nanomsg').socket('push', { dontwait: 1 } ); // async I/O

var pull = require('nanomsg').socket('pull', { dontwait: 0 } ); // socket blocks
```

@nickdesaulniers, review? ~~Should we also add a note about this in the readme?~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,7 @@ nano.socket('bus', { raw: true } );
 * `'rcvmaxsize'` *(Number, default: `false`)*: see [`socket.rcvmaxsize(size)`](https://github.com/nickdesaulniers/node-nanomsg#socketrcvmaxsizesize)
 * `'chan'` *(Array, default: `['']`)*: see [`socket.chan(Array)`](https://github.com/nickdesaulniers/node-nanomsg#socketchanarray)
 * `'wsopt'` *(String, default: `'binary'`)*: see [`socket.wsopt(str)`](https://github.com/nickdesaulniers/node-nanomsg#socketwsoptstr)
+* `'dontwait'` *(Number, default: `'1'`)*: see [`socket.dontwait(n)`](https://github.com/nickdesaulniers/node-nanomsg#socketdontwaitn)
 
 ### socket.shutdown(address)
 
@@ -334,6 +335,23 @@ console.log(socket.wsopt()); // 'text'
 ```
 
 If you are implementing nanomsg websockets in the browser, please carefully review the spec: https://raw.githubusercontent.com/nanomsg/nanomsg/master/rfc/sp-websocket-mapping-01.txt
+
+### socket.dontwait(n)
+
+*(Function, param: Number, one or zero, default: `1`, except PUSH sockets)*: Sets the NN_DONTWAIT flag, specifying that the operation should be performed in non-blocking mode,
+
+* `1` for non-blocking mode
+* `0` for blocking mode
+
+Pass no parameter for the socket's current mode.
+
+```js
+socket.dontwait(0);
+console.log(socket.dontwait()); // 0
+
+// or set when socket is started:
+require('nanomsg').socket('pub', { dontwait: 0 });
+```
 
 # test
 
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ nano.socket('bus', { raw: true } );
 * `'rcvmaxsize'` *(Number, default: `false`)*: see [`socket.rcvmaxsize(size)`](https://github.com/nickdesaulniers/node-nanomsg#socketrcvmaxsizesize)
 * `'chan'` *(Array, default: `['']`)*: see [`socket.chan(Array)`](https://github.com/nickdesaulniers/node-nanomsg#socketchanarray)
 * `'wsopt'` *(String, default: `'binary'`)*: see [`socket.wsopt(str)`](https://github.com/nickdesaulniers/node-nanomsg#socketwsoptstr)
-* `'dontwait'` *(Number, default: `'1'`)*: see [`socket.dontwait(n)`](https://github.com/nickdesaulniers/node-nanomsg#socketdontwaitn)
+* `'dontwait'` *(boolean, default: `true`)*: see [`socket.dontwait(boolean)`](https://github.com/nickdesaulniers/node-nanomsg#socketdontwaitboolean)
 
 ### socket.shutdown(address)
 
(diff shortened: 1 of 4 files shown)


reqshark:
added `dontwait` sockopt w/ note in readme:

### socket.dontwait(n)

*(Function, param: Number, one or zero, default: `1`, except PUSH sockets)*: Sets the NN_DONTWAIT flag, specifying that the operation should be performed in non-blocking mode,

* `1` for non-blocking mode
* `0` for blocking mode

Pass no parameter for the socket's current mode.

```js
socket.dontwait(0);
console.log(socket.dontwait()); // 0

// or set when socket is started:
require('nanomsg').socket('pub', { dontwait: 0 });
```

README.md
```diff
@@ -67,6 +67,7 @@ nano.socket('bus', { raw: true } );
 * `'rcvmaxsize'` *(Number, default: `false`)*: see [`socket.rcvmaxsize(size)`](https://github.com/nickdesaulniers/node-nanomsg#socketrcvmaxsizesize)
 * `'chan'` *(Array, default: `['']`)*: see [`socket.chan(Array)`](https://github.com/nickdesaulniers/node-nanomsg#socketchanarray)
 * `'wsopt'` *(String, default: `'binary'`)*: see [`socket.wsopt(str)`](https://github.com/nickdesaulniers/node-nanomsg#socketwsoptstr)
+* `'dontwait'` *(Number, default: `'1'`)*: see [`socket.dontwait(n)`](https://github.com/nickdesaulniers/node-nanomsg#socketdontwaitn)
```
nickdesaulniers: can we make it a boolean instead of a number with 2 relevant values?

lib/index.js
```diff
@@ -459,6 +463,18 @@ Socket.prototype.rmchan     = function() {
...
+Socket.prototype.dontwait   = function (n) {
+  if ( arguments.length ) {
+    if (typeof n !== 'number' || n > 1 || n < 0) {
+      throw new TypeError('dontwait option must be either 1 or 0')
+    }
+    this.nndontwait = n;
+    return true;
+  } else {
+    return this.nndontwait;
+  }
```
nickdesaulniers: this function now can return a number or bool.  Not that someone's going to call it a lot, but we should stick with one return type (bool).

test/pushstress.js
```diff
@@ -0,0 +1,48 @@
...
+    if (++i === 8) {
+      t.equal(push.nndontwait, 1);
+      t.equal(this.nndontwait, 0);
+      push.close();
+      pull.close();
+    }
+  });
+  while (i--)
+    push.send(msg);
+});
```
nickdesaulniers: documentation + tests in a PR? :tada:

test/pushstress.js
```diff
@@ -0,0 +1,48 @@
...
+  pull.connect(addr);
+  push.bind(addr);
+  pull.on('data', function(data){
+    if (data !== msg)
+      throw new Error('assertion failed');
+    if (++i === 9998) {
+      t.equal(push.nndontwait, 0);
+      t.equal(this.nndontwait, 1);
+      push.close();
+      pull.close();
```
nickdesaulniers: `this` and `pull` refer to the same object in this context.  Let's stick with one, not both (not using `this` reads quicker)

lib/index.js
```diff
@@ -459,6 +463,18 @@ Socket.prototype.rmchan     = function() {
   };
 }
 
+Socket.prototype.dontwait   = function (n) {
+  if ( arguments.length ) {
+    if (typeof n !== 'number' || n > 1 || n < 0) {
+      throw new TypeError('dontwait option must be either 1 or 0')
+    }
```
nickdesaulniers: if it's a bool than we don't need this conditional

lib/index.js
```diff
@@ -459,6 +463,18 @@ Socket.prototype.rmchan     = function() {
...
+    this.nndontwait = n;
+    return true;
+  } else {
+    return this.nndontwait;
+  }
+}
```
nickdesaulniers: this whole function could be:

```js
    return (this.nndontwait = !!n);
```

lib/index.js
```diff
@@ -54,6 +54,7 @@ function Socket (type, opts) {
             this.protocol = nn.NN_PUSH;
             this.sender = true;
             this.receiver = false;
+            opts.dontwait = opts.dontwait || 0;
```
reqshark: yea but only for PUSH

README.md
reqshark: yea i thought about making it a boolean and wasn't sure which to go with, so if you have an opinion, i'll definitely make this bool and update

lib/index.js
reqshark:
```js
  case 'push':
    this.protocol = nn.NN_PUSH;
    this.sender = true;
    this.receiver = false;
    opts.dontwait = opts.dontwait || 0;
```

lib/index.js
```diff
@@ -102,6 +103,9 @@ function Socket (type, opts) {
   this.connected = {};
   this.queue = [];
 
+  /* async I/O option */
+  this.nndontwait = 'dontwait' in opts ? opts.dontwait : nn.NN_DONTWAIT;
```
reqshark: no, see previous comment

lib/index.js
```diff
@@ -181,7 +185,7 @@ Socket.prototype._receive = function () {
     return;
   }
 
-  var msg = nn.Recv(this.binding, nn.NN_DONTWAIT);
+  var msg = nn.Recv(this.binding, this.nndontwait);
```
reqshark: updated to bool so we're not passing value from user anymore

test/pushstress.js
reqshark: 🍻

lib/index.js
```diff
@@ -459,6 +464,20 @@ Socket.prototype.rmchan     = function() {
   };
 }
 
+Socket.prototype.dontwait   = function (bool) {
+  if ( arguments.length ) {
```
nickdesaulniers: please remove spaces inside parens

reqshark: done!

nickdesaulniers:
thanks Bent 🍻

PR #178, opened 2017-07-10
Update index.js

xujif:
fix setsockopt method

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -311,11 +311,11 @@ Socket.prototype.send = function (buf, flags) {
 
 /* returns an int, a string, or throws EBADF, ENOPROTOOPT, ETERM */
 Socket.prototype.getsockopt = function (level, option) {
-    return this._protectArray(nn.Getsockopt(this.binding, level, option));
+    return this._protectArray(nn.Getopt(this.binding, level, option));
 };
 
 Socket.prototype.setsockopt = function (level, option, value) {
-    return this._protect(nn.Setsockopt(this.binding, level, option, value));
+    return this._protect(nn.Setopt(this.binding, level, option, value));
 };
 
 /**


lib/index.js
```diff
@@ -311,11 +311,11 @@ Socket.prototype.send = function (buf, flags) {
 
 /* returns an int, a string, or throws EBADF, ENOPROTOOPT, ETERM */
 Socket.prototype.getsockopt = function (level, option) {
-    return this._protectArray(nn.Getsockopt(this.binding, level, option));
+    return this._protectArray(nn.Getopt(this.binding, level, option));
 };
 
 Socket.prototype.setsockopt = function (level, option, value) {
-    return this._protect(nn.Setsockopt(this.binding, level, option, value));
+    return this._protect(nn.Setopt(this.binding, level, option, value));
```
nickdesaulniers: huh, I wonder when this broke? Did we not have tests for these methods?

nickdesaulniers:
The commit message needs to be updated with what the commit actually does, maybe something like:

```
Fix get/setsockopt methods to call correct native methods

broken by commit <sha>
```

nickdesaulniers:
Also, a test would be good to prevent this kind of error in the future.

reqshark:
our sockopt api doesn't call those methods, AFAIK we don't ever call these:

```js
Socket.prototype.getsockopt

Socket.prototype.setsockopt
```

that being said, if they worked they might be handy functions so the user could do their own `nn_setsockopt` and `nn_getsockopt` calls from javascript

reqshark:
hi @xujif, 

After sleeping on it and giving this a little thought, I think we'd like to remove these unused methods from `Socket` prototype. 

#### some history: 

These methods on our prototype were written before we implemented JavaScript [`sockopt` api](README.md#sockopt-api). 

Back then our intention was to remove and replace these but we never got around to it.

Calling various nanomsg socket options correctly can be dicey because the underlying C implementation is not consistent everywhere with respect to both `nn_setsockopt`/`nn_getsockopt`. Easing that burden and reducing its complexity is actually the whole reason behind our `sockopt` api

Can you please update PR and remove `getsockopt` and `setsockopt` from Socket prototype, also please delete the code comment above them!

If you found a nanomsg socket option in the C library that we missed, please defer to our implementation and submit the PR accordingly, otherwise feel free to request the feature 🍺   

Thanks!

reqshark:
@xujif, ping! 

would you care to update? 

if no response by end of august (45+ days), the PR police force will have to close and update

🚓 🚓 🚓  woop woop

PR #179, opened 2017-09-28
delete old/unused Socket methods

reqshark:
details for PR in commit with extra rambling and entertainment on what this code was

also we need to add support for TTL option:

> NN_TTL
Retrieves the maximum number of "hops" a message can go through before it is dropped. Each time the message is received (for example via the nn_device(3) function) counts as a single hop. This provides a form of protection against inadvertent loops.

http://nanomsg.org/v1.0.0/nn_getsockopt.3.html

also suspect #178 has been detained in the closed tab... 🚔

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ libtool
 nanocat
 .libs
 v8
+package-lock.json
--- a/lib/index.js
+++ b/lib/index.js
@@ -309,15 +309,6 @@ Socket.prototype.send = function (buf, flags) {
     return buf.length;
 };
 
-/* returns an int, a string, or throws EBADF, ENOPROTOOPT, ETERM */
-Socket.prototype.getsockopt = function (level, option) {
-    return this._protectArray(nn.Getsockopt(this.binding, level, option));
-};
-
-Socket.prototype.setsockopt = function (level, option, value) {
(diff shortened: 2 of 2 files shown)


nickdesaulniers:
Filed #180 for NN_TTL.

This method is publicly exposed, right?

I think it's hard to say that these methods are unused.  They're exposed as part of the public API, so this would be a breaking change that others may rely on, though they are not documented in the README.  Removal of a publicly exposed method would warrant a major version release, and this change doesn't meet the bar, IMO.

I think it doesn't hurt to leave it, and the maintence burden of 2 1-line functions is virtually zero.  I vote to keep the methods. ☮️ 

Do we have tests covering them? [I don't think so](https://github.com/nickdesaulniers/node-nanomsg/blob/master/test/sockoptapi.js).

reqshark:
> I think it doesn't hurt to leave it, and the maintence burden of 2 1-line functions is virtually zero. I vote to keep the methods

then if we're leaving these methods we need to fix them. At some point they stopped working so we would need to make them work again. 

are you suggesting re-openning and merging #178 ? that's the first step in keeping them.

nickdesaulniers:
ah, sorry about that, forgot all the context in #178.  In that case, removing a broken API is probably fine, still might want to consider a major version bump.

PR #182, opened 2017-10-29
deps/nanomsg: upgrade to nanomsg 1.1.0 Fixes #181

### LINKED ISSUE #181: nanomsg 1.1.0 is out
Can I please request a new release since nanomsg v1.1.0 has just been released.

Regards.
Deepak.

nickdesaulniers:
This commit has a couple of needed changes, along with the upgrade:
* remove nanomsg source files that have been deleted upstream from
  our build.
* change the use of fs.unlink to fs.unlinkSync to prevent deprecation
  warnings:

(node:4821) [DEP0013] DeprecationWarning: Calling an asynchronous
function without callback is deprecated.

* fix up mksymbols.rb.  Upstream commit
  327a29b1a614131697c4be5da599dcc8368fdf04 broke our parsing script.
* rerun ./test/mksymbols.rb > test/symbols.js. This is required per
  nanomsg upgrade.


nickdesaulniers:
DO NOT MERGE YET

there are 2 more issues to fix before merging:

- [ ]  test/sockopt.js has 2 failing tests where reading back the value of nn.linger is 0.
- [ ]  we get an undefined symbol nn_strcasestr warning.

nickdesaulniers:
[upstream commit ea1c775ea00e9eae6d229e0beadc03c2e53beba4 deprecated NN_LINGER](https://github.com/nanomsg/nanomsg/commit/ea1c775ea00e9eae6d229e0beadc03c2e53beba4)

nickdesaulniers:
the aborts are coming from libuv: https://github.com/libuv/libuv/blob/719dfecf95b0c74af6494f05049e56d5771ebfae/src/unix/linux-core.c#L188

nickdesaulniers:
note: the deprecation warnings and abort happen before this commit, so I could separate them out. or we could punt the term fix, since it's not introduced here.

nickdesaulniers:
I think we may need to leave linger; the symbol tests look it up.  I don't want to modify the symbol generation.

nickdesaulniers:
this isn't ready for review.  we keep working on it later.

reqshark:
nice start Nick!

test/connect.js
```diff
@@ -29,7 +29,7 @@ test('map connect address eid for valid IPC address', function (t) {
   }
 
   sock.close();
-  require('fs').unlink('some_address');
+  require('fs').unlinkSync('some_address');
```
reqshark: good call...

btw what a shame nodejs has done away with **async fire-and-forgets**

reqshark: also I have a feeling this might not need to be unlinked after the update... IPC sockets ought to clean up after themselves, and if not then either libnanomsg or node-nanomsg should probably do that. Imagine if you had to delete TCP ports after using them. 

Right.. I can't remember whether OSX or Linux clears out that descriptor afterwards, but I think at least one of them leaves it behind, 

which means `unlink()` might do like an `ENOENT` or something (perfect use case for **async fire-and-forget**)

nickdesaulniers:
I think rather than remove linger, we should change the test to read back 0 always, as the underlying lib does.

nickdesaulniers:
The test/standalone/term.js test fails before this. Let's file an issue and follow up.

nickdesaulniers:
Ready for review @reqshark

nickdesaulniers:
> looks like sockopt ipv6 test is failing..

I suspect the test is flaky; I can run it locally np, maybe travis has issues with binding to certain ipv6 ports.  For example, multiple previous runs of this PR passed np.

test/mksymbols.rb
```diff
@@ -5,33 +5,29 @@
...
-        if a.length == 5
-            opts.push(<<-FOO)
-                'value': nn.#{a[0]},
-                'name':  #{a[1].strip},
-                'ns':    nn.#{a[2]},
-                'type':  nn.#{a[3].strip},
-                'unit':  nn.#{a[4]}
-            FOO
-        end
+    m[1].scan(/NN_SYM\((\w+),\s*(\w+),\s*(\w+),\s*(\w+)\),?/) do |n|
```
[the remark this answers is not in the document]
nickdesaulniers: Ruby's scan is just a foreach over a group of matches.  Ruby in general has some pretty slick functionality in its std lib.

reqshark:
> I suspect the test is flaky;

ya let's get rid of it then and we can add it back later if we figure out what's going on there w/ travis

nickdesaulniers:
> ya let's get rid of it then and we can add it back later if we figure out what's going on there w/ travis

is there a way to rekick travis? I'd rather see if we can get a pass and keep the test.

nickdesaulniers:
something is weird with travis: look at these alternating reds: https://travis-ci.org/nickdesaulniers/node-nanomsg/requests

reqshark:
ya i just kicked let's see. probably that TCP socket option thing for ephemeral port reuse

reqshark:
if nanomsg had a default linger before, makes sense the port might not be free for reuse binding... 

lingering around might have given the travis stack/kernel enough to make port 6000 available again in time for the next `bind()`

just kicked off a build to confirm, but thinking pick another port other than 6000 for that test or figure out a way to set `SO_REUSEADDR` socket option on nanomsg sockets

reqshark:
> there are 3 more issues to fix before merging:

> - [x]  test/sockopt.js has 2 failing tests where reading back the value of nn.linger is 0.
> - [x]  we get an undefined symbol nn_strcasestr warning.
> - [x] tests/standalone/term.js now aborts

i marked item 3 as checked off since that got fixed in the other PR and ipv6 is passing on a regular linux, I'll make a Travis issue to fix that and I think this is good to merge

PR #184, opened 2017-11-04
stop libuv polling on every socket on calls to term Fixes #183

### LINKED ISSUE #183: tests/standalone/term.js is failing
we should fix this.  I probably regressed this in https://github.com/nickdesaulniers/node-nanomsg/commit/0461f5c20cf83865d22235787e48860e760c7e39.

It's a stupid test, so not high priority.

nickdesaulniers:


nickdesaulniers:
maybe want some way to remove them from the global array that way they can be GCed?

nickdesaulniers:
but after a call to term, nanomsg effectively stops working, so it's probably fine to leak these socket objects.

PR #185, opened 2017-11-05
check on ipv6

reqshark:
dont merge! 

PRing to kick off a build

out of curiosity from intermittent ipv6 test fails


reqshark:
wow travis is killing us... why does it take so long to run a test? that webhook fired like 10 mins ago

reqshark:
nevermind we're missing a new nanomsg option

Issue #206, opened 2019-06-25
Compile error in Node V12.*


alkan:
We get the following error while installing the package>

`gyp ERR! build error 
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/home/ayhan/devtools/node-12.4.0/lib/node_modules/npm/node_modules/node-gyp/lib/build.js:262:23)
gyp ERR! stack     at ChildProcess.emit (events.js:200:13)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:272:12)
gyp ERR! System Linux 4.15.0-52-generic
gyp ERR! command "/home/ayhan/devtools/node-12.4.0/bin/node" "/home/ayhan/devtools/node-12.4.0/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /home/ayhan/projects/matriks/dev/work/nnw/node_modules/nanomsg
gyp ERR! node -v v12.4.0
gyp ERR! node-gyp -v v3.8.0
gyp ERR! not ok 
n`

nickdesaulniers:
thanks for the report @alkan . Can you provide the full error logs? Otherwise the report provided just looks like `gyp ERR!` repeated again and again without any meaningful errors with which to go on.

Also, how did you invoke the build?

enjoy05sky:
Use Vue-Cli install nanomsg4.0.2 .My nodejs version is v12.0.0

``..\src\node_nanomsg.cc(102): error C2664: 'v8::String::Utf8Value::Utf8Value(const v8::String::Utf8Value &)': cannot convert argument 1 from 'v8::Local<v8::Value>' to 'const v8::String::Utf8Value &' [F:\ali-project\Kratos\node_modules\nanomsg\build\node_nanomsg.vcxproj]
  ..\src\node_nanomsg.cc(102): note: Reason: cannot convert from 'v8::Local<v8::Value>' to 'const v8::String::Utf8Value'
  ..\src\node_nanomsg.cc(102): note: No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called
[...]

PR #207, opened 2019-06-30
.travis.yml: add node 12 builds

nickdesaulniers:
Bug #206


nickdesaulniers:
this should be landed AFTER https://github.com/nickdesaulniers/node-nanomsg/pull/211 at which point I should rekick the build.

reqshark:
we might want to remove `v0.10`

nickdesaulniers:
> we might want to remove v0.10

Why? It still works.

Issue #214, opened 2019-07-01
remove package-lock.json from .gitignore


nickdesaulniers:
I'm seeing this when playing around with dependency updates:
```
npm notice created a lockfile as package-lock.json. You should commit this file.
```
looks like it's in our `.gitignore`. https://github.com/nickdesaulniers/node-nanomsg/blob/60de923f07e7bee906a0e2a8509278648e163a3c/.gitignore#L17

We should remove that and commit this; it's important for package integrity checks.

nickdesaulniers:
we should do this AFTER landing #212 an #213 .

nickdesaulniers:
fixed in https://github.com/nickdesaulniers/node-nanomsg/commit/87411d4b750008185ad5e34dda7f8f541d96c4b3

PR #219, opened 2019-07-22
Remove package-lock.json from .gitignore

vvscode:
https://github.com/nickdesaulniers/node-nanomsg/issues/214


nickdesaulniers:
thanks for the patch! I'm not able to test just yet; but one of the maintainers will need to verify the package-lock.json manually before reviewing.

Issue #224, opened 2020-11-27
(node:660) electron: The default of contextIsolation is deprecated and will be c hanging from false to true in a future release of Electron.  


dnybz:
Hi, i am use electron 11.0.3  (Node.js 12.18.3, Chromium 87.0.4280.67, and Electron 11.0.3)

run app errors:

(node:660) electron: The default of contextIsolation is deprecated and will be c
hanging from false to true in a future release of Electron.  See https://github.
com/electron/electron/issues/23506 for more information
(node:13928) Electron: Loading non-context-aware native module in renderer: '\\?
\G:\workspace\project\trunk\src\electron-client\appclient\node_modules\nanomsg\bui
ld\Release\node_nanomsg.node'. This is deprecated, see https://github.com/electr
on/electron/issues/18397.

https://github.com/nickdesaulniers/node-nanomsg/blob/master/src/node_nanomsg.cc#L286

//NODE_MODULE(node_nanomsg, InitAll)
#if NODE_MAJOR_VERSION >= 10
NAN_MODULE_WORKER_ENABLED(node_nanomsg, InitAll)
#else
NODE_MODULE(node_nanomsg, InitAll)
#endif

rebuild ok ~

Bartel-C8:
So @dnybz , how did you solve this?

Bartel-C8:
Actually, your solution solved it. I was just rebuilding it wrongly (with yarn rebuild / npm rebuild). But I was using electron-rebuild, which was/should be responsible of rebuilding the native module correctly (with the correct (ABI) version of Electron).

Thanks!

PR #233, opened 2022-08-11
Support Electron > 14

Bartel-C8:
Use updated NAN_MODULE_WORKER_ENABLED macro from nan (which takes care of nodejs versioning)
https://github.com/nodejs/nan/commit/b058fb047d18a58250e66ae831444441c1f2ac7a

See electron deprecation issue:
https://github.com/electron/electron/issues/18397

Solves https://github.com/nickdesaulniers/node-nanomsg/issues/224

### DIFF
--- a/src/node_nanomsg.cc
+++ b/src/node_nanomsg.cc
@@ -283,4 +283,4 @@ NAN_MODULE_INIT(InitAll) {
   }
 }
 
-NODE_MODULE(node_nanomsg, InitAll)
+NAN_MODULE_WORKER_ENABLED(node_nanomsg, InitAll)


Bartel-C8:
@nickdesaulniers : could you review it please? As nan has been recently upgraded, this should cause no problems.

nickdesaulniers:
Sorry for the delay; have been sick with Covid; still am.

Do we need to upgrade the version of NAN used, too?
https://github.com/nickdesaulniers/node-nanomsg/blob/f339119585797bebebb17fc02028db61904ba783/package.json#L8

Bartel-C8:
Take care!

No, no upgrade of nan is needed as 2.15.0 and above contains the mentioned patch which is needed: https://github.com/nodejs/nan/commit/b058fb047d18a58250e66ae831444441c1f2ac7a

nickdesaulniers:
thanks for the patch, I will work on publishing a new release tonight

nickdesaulniers:
published https://www.npmjs.com/package/nanomsg/v/4.2.1

PR #237, opened 2023-08-05
can filter with Buffer chan

### LINKED ISSUE #236: sub: filter by binary chan prefix?
I'm transfering all messages between nanomsg nodes in binary format. But filtering with `chan` option do not work:
```js
let chan = ["\x95"] // tried Buffer.from([0x95]) too
let sub = socket('sub', {chan})
// sub doesn't receive messages starting with [0x95, ...]
```
Look like because chan converted to utf-8 string before passing to `nn_setsockopt`
    https://github.com/nickdesaulniers/node-nanomsg/blob/428455eae354c44a6ef77a7d24ab82b3757bfc60/src/node_nanomsg.cc#L59

farwayer:
fix #236

### DIFF
--- a/src/node_nanomsg.cc
+++ b/src/node_nanomsg.cc
@@ -57,15 +57,21 @@ NAN_METHOD(Chan) {
   int level = NN_SUB;
   int option = Nan::To<int>(info[1]).FromJust();
 
+  void *value;
+  size_t len;
+
   if (node::Buffer::HasInstance(info[2])) {
-    info.GetReturnValue().Set(Nan::New<v8::Number>(nn_setsockopt(
-        s, level, option, node::Buffer::Data(info[2]),
-        node::Buffer::Length(info[2]))));
+    value = node::Buffer::Data(info[2]);
+    len = node::Buffer::Length(info[2]);
   } else {
     Nan::Utf8String str(info[2]);
-    info.GetReturnValue().Set(Nan::New<v8::Number>(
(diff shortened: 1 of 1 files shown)


src/node_nanomsg.cc
```diff
@@ -56,10 +56,16 @@ NAN_METHOD(Chan) {
...
-  info.GetReturnValue().Set(Nan::New<v8::Number>(
+  if (node::Buffer::HasInstance(info[2])) {
+    info.GetReturnValue().Set(Nan::New<v8::Number>(nn_setsockopt(
+        s, level, option, node::Buffer::Data(info[2]),
+        node::Buffer::Length(info[2]))));
+  } else {
+    Nan::Utf8String str(info[2]);
+    info.GetReturnValue().Set(Nan::New<v8::Number>(
         nn_setsockopt(s, level, option, *str, str.length())));
+    }
```
nickdesaulniers, line 68: There's a fair amount of duplication on either side of that condition.  Perhaps you can call `nn_setsockopt` two different ways in either branch, but then have the common tail construct the v8::Number and call Set after?

nickdesaulniers (commented):
Thanks for the patch!

src/node_nanomsg.cc
farwayer: I used the same code style as in Send method but sure!

nickdesaulniers (approved):

